PR #20, opened 2023-11-01
Update deps and remove mix and match of jakarta.servlet and javax.servlet.

MikeNeilson:
Initial cleanup of dependencies. Including referencing the opendcs.jar that's in maven central.

I noticed that there were references to both javax.servlet and jakarta.servlet and these should not be mixed. Needs to be all or nothing.

Given that we want the option of Tomcat, up to version 9, we should stick with javax.servlet for the moment.

However, if we start targeting java11 vs java 8, it may make more sense to make the jump.

I also took the liberty of updating the dependencies that were listed.

NOTE: the only testing done was making sure the API started. Still need to verify I can login and get anything.


wjonassen (approved):
Looks good.  I will see if I can test it out tonight.

MikeNeilson:
I was able to verify things still working, at least for one endpoint, on my end.

MikeNeilson:
@wjonassen let me know how you want to handle merge changes in at that point in time. If you aren't used to working with someone else on git and can be a little confusing when you suddenly can't push changes because something was already merged in.

wjonassen:
> @wjonassen let me know how you want to handle merge changes in at that point in time. If you aren't used to working with someone else on git and can be a little confusing when you suddenly can't push changes because something was already merged in.

@MikeNeilson I say merge them in.  I work with others on git, so I have dealt with the hardships of merging conflicting code.  That said, we should have a chat on any rules/preferences for merging code that we want to follow.

I am out of the office right now, coming back Thursday.  If you have any time Thurs/Fri, or any time next week, we could do a catch up to discuss this, as well as where to focus our time on the API/Client.

MikeNeilson:
awesome. the next bit I'm going to work on is putting the token in a header + using the standard SecurityContext mechanism, but I'll leave it working in the header so it's not a huge change to the client yet.

PR #64, opened 2023-12-08
Add OWASP Zap API scanner to CI/CD

adamkorynta:
## Problem Description

SonarCloud only provides a static analysis of security vulnerabilities. A runtime analysis is also needed.

## Solution

Start docker containers with the Postgres database and jetty installation and run the OWASP Zap automated tool from another docker container against the OpenDCS API.

## how you tested the change

Testing docker containers locally and from GitHub Actions. 
The GitHub summary shows the summary of the Markdown output report. The HTML report is attached as a build artifact.

## Where the following done:

- [X] Was relevant documentation updated?


gradle.properties
```diff
@@ -1,2 +1,5 @@
 #DCSTOOL_HOME
```
MikeNeilson, line 1: this should be a `gradle.properties.example` with `gradle.properties` excluded from the build.
Less risk of credential spillage that way.

odcsapi-jetty-standalone/build.gradle
```diff
@@ -34,15 +35,53 @@ publishing {
...
+    def username = project.findProperty("db.username")
+    def password =  project.findProperty("db.password")
+    def file = project.file("${project.buildDir}/opendcs/credentials.txt")
+    file.write("${username}\n${password}\n")
+}
+
+task writeUserAuthFile(type: JavaExec) { ->
+    dependsOn writeCredentials
+    classpath = sourceSets.test.runtimeClasspath
+    mainClass = "ilex.util.UserAuthFile"
```
MikeNeilson, line 67: FYI opendcs 7.0.5+ have support for retrieving credentials from the environment

use AuthSource.getCredentials(<the db auth file string>);

See https://github.com/opendcs/opendcs/blob/2fc798892dc1d49840e579e75ff0ca43bc72c54d/src/test/java/org/opendcs/authentication/EnvironmentAuthSourceTest.java#L27 for a full examples

odcsapi-jetty-standalone/config/user.properties
```diff
@@ -0,0 +1,106 @@
...
+siteNameTypePreference=NWSHB5
+pyTsRoleColor=0x8B0000
+decwizRawDataDir=$HOME/raw-done
+retryFailedComputations=false
+compCheckPeriod=120
+showComputationEditor=true
+showPlatformMonitor=true
+showRoutingMonitor=true
+showNetlistEditor=false
+DbAuthFile=$DCSTOOL_USERDIR/.decodes.auth
```
MikeNeilson, line 48: this line can be changed to the EnvAuth spec as in the example above.

MikeNeilson:
FYI, I'm working in this PR specifically for this project: https://github.com/opendcs/opendcs/pull/387

You might want to hold off on some of this. The above project (which already exists, that PR is just making usable artifacts) already handles database initialization and initial dataload.

It's not unreasonable at this stage to inform the dev that they need an existing database setup and we're working on some automation.

adamkorynta:
Currently the initial database table load isn't working right and the scripts need a couple of hacks to get the schema to install correctly. I'll hold off on anymore improvement to that until after your opendcs PR is merged.

Additionally, the OWASP Zap docker scan is occurring locally, but the swagger.json isn't auto-generated so the "odcsapi" endpoint winds up being relative to the swagger.json endpoint (resulting in scans of funny endpoints like `"http://opendcs:8000/swagger.json/odcsapi/reflists"`) so I will have to pause on this effort until we can get the swagger.json autogenerated.

MikeNeilson:
This PR: https://github.com/opendcs/opendcs/pull/433 will probably be more appropriate.
I'm still working out issues with data in classpaths and such (trying to keep it clean to the downstream programs)

But this will certainly make the actual schema install easier.

odcsapi-jetty-standalone/build.gradle
adamkorynta: Perfect, thanks!

wjonassen (approved):

opendcs-rest-api/src/main/resources/swaggerui/open_api.json
```diff
@@ -6,9 +6,6 @@
     "version": "0.0.3"
   },
   "servers": [
-    {
-      "url": ""
```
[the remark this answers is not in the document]
adamkorynta: yeah, I believe the "servers" section should be completely removed for that reason, I just don't know how it automatically gets picked up.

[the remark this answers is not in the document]
adamkorynta: I added a comment to: https://github.com/opendcs/rest_api/issues/78

PR #80, opened 2023-12-16
update contributing.md to point to sonarcloud rules

adamkorynta:
## Problem Description

SonarCloud rules are somewhat buried within the OpenDCS organization.

## Solution

Adding a link to the contributing.md to point to where the rules are configured.

## how you tested the change

Tested link.

## Where the following done:

- [X] Was relevant documentation updated?


CONTRIBUTING.md
```diff
@@ -123,9 +123,11 @@ The current java target is java 8. Please make sure you do not add java 9+ featu
 
 Now you're ready to [clone the repository](https://help.github.com/articles/cloning-a-repository/) and start looking at things. If you are going to submit any changes you make, please read the Submitting changes section below.
 
-### Code Style
+### Coding Practices
 
-TODO: add checkstyle to gradle build
+Code style rules as well as other static analysis rules are setup for the [OpenDCS Organization on SonarCloud](https://sonarcloud.io/organizations/opendcs/quality_profiles/show?name=CWMS_ATO&language=java).
```
MikeNeilson, line 128: I assume we're going to change the quality profile to something "OpenDCS" and adjust the rules accordingly at some point?

MikeNeilson (approved):

CONTRIBUTING.md
adamkorynta: Updated to just "OpenDCS". We can always extend/rename later to create an "OpenDCS REST API" in the future as needed

MikeNeilson (commented):
You can delete the .DS_store files, I think.
If those folders are empty and need to stay then something like ".keep" is better.
But I'm pretty sure .DS_store is from CSV or SVN

adamkorynta:
.DS_store are MacOS only. They were excluded in the .gitignore so they didn't get purged when I reorganized the directories.

PR #87, opened 2024-01-10
Remove insecure printStackTrace calls.

### LINKED ISSUE #96: Remove insecure printStackTrace calls.
**Describe the bug**
There are several printStackTrace calls in the code.  Throwable.printStackTrace(...) prints a Throwable and its stack trace to System.Err (by default) which is not easily parseable and can expose sensitive information.

**To Reproduce**
You can see in the code that these problems exist.

**Expected behavior**
Instead of printstacktrace, use a logger to show the message.

wjonassen:
## Problem Description
There are several calls to throwable.printStackTrace throughout the rest-api project.  These have been flagged as insecure and need to be removed.
Fixes #96. 
Throwable.printStackTrace(...) prints a Throwable and its stack trace to System.Err (by default) which is not easily parseable and can expose sensitive information.

## Solution
Use a logger to get the same message to the user.

## how you tested the change
Forced one of these conditions and verified the logger logged the message.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/lrgsclient/ApiLddsClient.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/lrgsclient/ApiLddsClient.java
@@ -537,8 +537,6 @@ debug(module + ".sendAuthHello authenticator '" + authStr + "'");
 		{
 			String em = "Error in LrgsStatusXio: " + ex;
 			warning(em);
-			System.err.println(em);
-			ex.printStackTrace();
 			throw new DdsProtocolError(em);
 		}
 	}
@@ -602,8 +600,6 @@ debug(module + ".sendAuthHello authenticator '" + authStr + "'");
 		{
 			String em = "Error in RawMessageBlockParser: " + ex;
 			warning(em);
-			System.err.println(em);
-			ex.printStackTrace();
 			throw new DdsProtocolError(em);
 		}
 	}
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/lrgsclient/ApiLddsClient.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/lrgsclient/ApiLddsClient.java
@@ -27,8 +27,9 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.StringTokenizer;
 import java.util.TimeZone;
-import java.util.logging.Level;
-import java.util.logging.Logger;
+
+import org.slf4j.LoggerFactory;
+
 import java.util.zip.GZIPInputStream;
 
 import org.opendcs.odcsapi.beans.ApiLrgsStatus;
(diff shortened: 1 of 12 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -308,9 +309,9 @@ public void procFinished(String procName, Object obj, int exitStatus)
 		}
 		catch (IOException ex)
 		{
-			ex.printStackTrace();
-			throw new WebAppException(ErrorCodes.DATABASE_ERROR,
-				"Error attempting to start appId=" + appId + ": " + ex);
+			Logger.getLogger(ApiConstants.loggerName).log(Level.WARNING,
```
adamkorynta, line 312: Add the exception to the WebAppException overload and remove the redundant log message

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/lrgsclient/ApiLddsClient.java
```diff
@@ -656,17 +651,17 @@ public void setStrongOnly(boolean strongOnly)
 
 	public void info(String str)
 	{
-		Logger.getLogger(ApiConstants.loggerName).info("LddsClient INFO: " + str);
+		LoggerFactory.getLogger(ApiConstants.loggerName).info("LddsClient INFO: {}", str);
```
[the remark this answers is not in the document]
adamkorynta: also it should be initialized as a constant in the class
`private static final LoggerFactory LOGGER = LoggerFactory.getLogger(ApiLddsClient.getClass());`

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/util/ApiBasicClient.java
```diff
@@ -143,30 +145,27 @@ public void disconnect( )
...
 			{
 				if (socket != null)
 					socket.close();
 			}
-			catch (Exception e)
+			catch (IOException ex)
 			{
-				System.out.println("Error closing socket: "+e.getMessage());
-				e.printStackTrace();
+				throw new IOException( "Error closing socket: ", ex);
```
adamkorynta, line 168: This doesn't add much context, I'd just remove the catch since it's throwing the same exception anyways

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/util/ApiBasicClient.java
```diff
@@ -143,30 +145,27 @@ public void disconnect( )
...
+            try
 			{
-				if (output != null)
-					output.close();
-			}
-			catch (Exception e)
+                if (output != null)
+                    output.close();
+            }
+			catch (IOException ex)
```
adamkorynta, line 157: I think we still want to close the socket even if the output.close() fails

wjonassen: I did update this in the new commit, the new code is much more readable, but technically if one close fails, the others won't be run.

Readability wise, it might make sense to do it this way?

adamkorynta (approved):

MikeNeilson (approved):

PR #90, opened 2024-01-15
SwaggerUI needs to run properly on Tomcat using a WAR file.

### LINKED ISSUE #89: Need to get Swagger UI working properly using Tomcat/WAR files.
Swagger UI Works using the jetty implementation, but it does not work when running on Tomcat using a WAR file.

Create a resource file, the same way the rest of the API works, and in the same location, to make the swagger client files available to the client.

wjonassen:
## Problem Description

Currently cannot run SwaggerUI on Tomcat using a WAR file.
Fixes #89. 

## Solution

Make the Swagger UI client files available to the web browser via a resource file.  Also, the SwaggerUI relative url path was moved to /swaggerui.

## how you tested the change

1. Loaded the swaggerui webpage.  Verified the page loaded properly.
       Opened developer tools in the web browser and verified there were no errors loading any of the resource files for swagger.
       Verified the base swagger file was loaded on page load.
       Verified a secondary swagger.json file can be loaded when the user changes the target swagger json file.
2. Ran a get against the database successfully.
3. Ran a post against the database successfully.

## Where the following done:

- [x] Was relevant documentation updated?


MikeNeilson:
This seems like a lot of manual work, something like this could also work:
https://github.com/USACE/cwms-data-api/blob/5dd6fbd7cb289368aa492c9547fe0f447c0138f1/cwms-data-api/build.gradle#L150-L179

There's still a need to maintain a few of the swagger files outside of version control for customizations but seems better than keeping the entire thing in version control.

opendcs-rest-api-jetty/build.gradle
```diff
@@ -46,6 +46,21 @@ task startJetty(type: JavaExec) { ->
             "-s")
 }
 
+task copySwagger(type: Copy) {
```
adamkorynta, line 49: I'm not sure what this task is doing, given the war/jar for the opendcs-rest-api already include these resources. Also this looks to be copying into a tmp directory, but I don't see where that tmp directory is used.

gradle/libs.versions.toml
```diff
@@ -9,13 +9,15 @@ slf4j = "2.0.10"
...
+swagger-ui = "5.10.3"
 glassfish-jaxb = "2.3.3"
 websocket = "1.1"
 javax-validation = "2.0.1.Final"
+
 #Test Dependencies
 junit = "5.10.1"
 mockito = "3.12.4"
-
+rest-assured = "2.9.0"
```
adamkorynta, line 20: This looks unused

gradle/libs.versions.toml
```diff
@@ -34,14 +36,18 @@ slf4j-jdk = { module = "org.slf4j:slf4j-jdk14", version.ref = "slf4j" }
...
+# webjars
+swagger-ui = { module = "org.webjars:swagger-ui", version.ref="swagger-ui" }
+
 #Test Dependencies
 junit-bom = { module = "org.junit:junit-bom", version.ref = "junit" }
 junit-api = { module = "org.junit.jupiter:junit-jupiter-api" }
 junit-engine = { module = "org.junit.jupiter:junit-jupiter-engine" }
 mockito-core = { module = "org.mockito:mockito-core", version.ref = "mockito" }
-mockito-junit-jupiter = { module = "org.mockito:mockito-junit-jupiter", version.ref = "mockito" }
+mockito-junit-jupiter = { module = "org.mockito:mockito-junit-jupiter", version.ref = "mockito" }
```
adamkorynta, line 53: This looks unused

opendcs-rest-api/build.gradle
```diff
@@ -14,8 +18,37 @@ dependencies {
     implementation(libs.json.jackson)
     implementation(libs.jersey.container.servlet)
     implementation(libs.jersey.hk2)
-    runtimeOnly(libs.jaxb.runtime)
+    implementation(libs.jaxb.runtime)
```
adamkorynta, line 21: The runtime jar shouldn't be on the compile-time class path

opendcs-rest-api/src/main/resources/swaggerui/open_api.json
```diff
@@ -7,10 +7,7 @@
   },
   "servers": [
     {
-      "url": ""
-    },
-    {
-      "url": "odcsapi"
+      "url": "."
```
adamkorynta, line 10: Maybe this will solve my issue with the OWASP Zap. Previously it was trying to make all of the endpoints relative to the swagger.json. I had to remove the empty URL, but maybe this will work better. Let's see what the GitHub Actions do once you rebase/merge with main

adamkorynta (changes requested):

adamkorynta (approved):

PR #114, opened 2024-01-23
Move context.xml to an examples directory.

### LINKED ISSUE #113: Move context.xml to an examples directory
**Component to be refactored**
Right now, context.xml has information in it that needs to be scrubbed, plus it's really an example file, so it should be moved to an example directory.

**Purpose of refactoring**
This file is an example file.  It belongs in a directory for examples.

**Proposed method**
Create an examples directory and move the file into it.  This directory can house more examples.

wjonassen:
## Problem Description
context.xml should be in an examples directory and scrubbed to have example information in it only.  It should also describe what the file does and how it's used.

Fixes #113. 

## Solution
This file has example information that is useful to the user.  Let's move it to an examples folder.
Describe your solution.

## How it was tested
Used this example context.xml on an actual implementation and the rest-api connected to the database.

## Where the following done:


opendcs-web-client/src/main/webapp/conf/api.conf
```diff
@@ -1,3 +1,3 @@
 url=http://127.0.0.1
-port=8081
-context=odcsapi
\ No newline at end of file
+port=8080
+context=opendcs-rest-api-99.feature-dynamic-sql-SNAPSHOT
```
adamkorynta, line 3: should probably leave this as odcsapi to match the word doc documentation

wjonassen: I didn't mean to add this change in.  I will revert that.

adamkorynta: yeah that makes sense

adamkorynta (approved):

adamkorynta (approved):

PR #121, opened 2024-02-14
Addressing the system.out lines throughout the project.

### LINKED ISSUE #79: Replace System.out.println references with SLF4j
**Component to be refactored**
Replace calls to System.out.println to use JUL logging instead.

**Purpose of refactoring**
Using Java Util Logging is more maintainable than using `System.out.println` for several reasons. Firstly, Java Util Logging provides different logging levels (SEVERE, WARNING, INFO, etc.), which allows for more granular control over what information is logged and when. This is not possible with `System.out.println`, which outputs everything regardless of its importance. Secondly, Java Util Logging supports output to different targets, not just the console, but also file systems, sockets, etc., and can be easily configured to change the output target. Thirdly, Java Util Logging supports formatting of log messages, which can make the logs easier to read and parse. Lastly, using a logging framework like Java Util Logging can improve performance, as it can be configured to log messages asynchronously. On the other hand, using `System.out.println` can slow down an application because it's a synchronous operation.

**Proposed method**
[SonarCloud](https://sonarcloud.io/project/issues?impactSoftwareQualities=MAINTAINABILITY&resolved=false&rules=java%3AS106&id=opendcs_rest_api) has identified quite a few calls that need to be rerouted . They should all be replaced, though the effort might be too high to do at once so this might be accomplished in several efforts or when surrounding code is updated.

**Additional context**
As a stopgap HEC has been using Flogger (Google's fluent logger API) to reroute all `System.out` calls to JUL and whenever code lines are updated surrounding the `System.out` code, they are replaced with JUL calls.

wjonassen:
## Problem Description
There are system.out's in the project.  We want to either remove them completely (if they are not useful), or replace those with SLF4J in the case that they are useful.

Fixes #79. 

## Solution
In the cases where it's a system.out that's not useful (maybe it's a debug line), remove it completely.  If it is useful, then convert it into slf4j.

## how you tested the change
Verify the project builds.

## Where the following done:


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ObjectMapperContextResolver.java
```diff
@@ -41,11 +46,13 @@ public ObjectMapper getContext(Class<?> type)
...
     {
-    	System.out.println("createObjectMapper");
-        ObjectMapper mapper = new ObjectMapper();
-        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'[z]");
-        sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
-        mapper.setDateFormat(sdf);
-        return mapper;
+        String dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'[z]";
+        String timezone = "UTC";
+        LOGGER.info("Creating object map using date format '{}' and timezone '{}'", dateFormat, timezone);
```
wjonassen, line 51: I am not sure what detail of logging is desired here.  I didn't fully think the system.out was necessary, but I do want to add some logging here.  Let me know if this type of logging is desired and I will keep it.  Otherwise, I can remove it.

Not sure if logging the dateformat and timezone is necessary either.  I can remove it if it's too much.

MikeNeilson: Honestly probably nothing unless there's an error. That particular line of code is going to get executed so often (basically for everything) that it would be difficult for all of us to not know it's correct.

wjonassen: Sounds good.  In that case I will just remove the system.out.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/appmon/ApiEventClient.java
```diff
@@ -29,39 +29,42 @@
...
 	public ApiEventClient(Long appId, String hostname, int port, String appName, long pid)
 	{
 		super(hostname, port);
 		this.appId = appId;
 		this.appName = appName;
 		sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
 		this.pid = pid;
-System.out.println("ApiEventClient appId=" + appId + ", host=" + hostname + ", port=" + port + ", name=" + appName);
-super.debug = System.out;
+		LOGGER.debug("ApiEventClient appId={}, host={}, port={}, name={}", appId, hostname, port, appName);
```
wjonassen, line 53: Converting these all from system.out debug statements to use slf4j.  They seem like useful debug statements, but let me know if they should not be in here and I can remove them.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/appmon/ApiEventClient.java
```diff
@@ -29,39 +29,42 @@
...
-	
 	public ApiEventClient(Long appId, String hostname, int port, String appName, long pid)
 	{
 		super(hostname, port);
 		this.appId = appId;
 		this.appName = appName;
 		sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
 		this.pid = pid;
-System.out.println("ApiEventClient appId=" + appId + ", host=" + hostname + ", port=" + port + ", name=" + appName);
-super.debug = System.out;
```
wjonassen, line 58: No longer need this, as we are moving to slf4j.

wjonassen:
@MikeNeilson I have addressed a bunch of the system.out's in this.  There are still more, but if you could go through some of the changes to make sure I'm addressing them appropriately, I can then finish off the rest of them.

I am removing the ones where they really shouldn't be there, converting some to slf4j logging, and converting some to thrown exceptions.

MikeNeilson (approved):
Looks like a reasonable start to me.

wjonassen:
> Looks like a reasonable start to me.

@MikeNeilson I know you did approve this yesterday, but I have just added the rest of the updates for addressing the system.out's and don't want to merge until the new changes are approved (or updates are needed).

MikeNeilson (approved):

PR #135, opened 2024-04-02
Give the OpenDcsDataTable class the ability to add cell borders.

wjonassen:
## Problem Description
Right now, data tables cannot have cell borders.  We need to have the ability to add in cell borders.

## Solution
Give the ability, from an OOP perspective, for datatables to have cell borders.  This will be an optional object that will be passed into the OpenDcsDataTable constructor, where by default the cell borders will be set to true if this optional object is not passed in.

## how you tested the change
Verified the cell borders exist if:
1) The object is passed in with cellBorders = true.
2) The object is not passed in.
Verified the cell borders do not exist if:
1) The object is passed in with cellBorders = false;

## Where the following done:

### DIFF
--- a/opendcs-web-client/src/main/webapp/resources/js/datatables/datatables.js
+++ b/opendcs-web-client/src/main/webapp/resources/js/datatables/datatables.js
@@ -26,7 +26,8 @@ class OpenDcsDataTable {
 	  }
     this.domId = domId;
     this.jq = $(`#${domId}`);
-    this.tableProperties = tableProperties != null ? tableProperties : defaultTableProperties;
+    this.tableProperties = tableProperties != undefined && tableProperties != null
+        ? tableProperties : defaultTableProperties;
     this.properties = properties;
     
     this.setInlineOptions(inlineOptions);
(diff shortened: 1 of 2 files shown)


opendcs-web-client/src/main/webapp/resources/js/datatables/datatables.js
```diff
@@ -14,16 +14,19 @@
...
+                                  };
 	  //Remove # if the passed id starts with one.
 	  if (domId.startsWith("#"))
 	  {
 		  domId = domId.slice(1);
 	  }
     this.domId = domId;
     this.jq = $(`#${domId}`);
-    
+    this.tableProperties = tableProperties != null ? tableProperties : defaultTableProperties;
```
MikeNeilson, line 29: i think needs needs an undefined check instead of or in addition to a null check.

MikeNeilson: But it's been a few months since I've done javascript.

MikeNeilson (approved):

opendcs-web-client/src/main/webapp/resources/js/datatables/datatables.js
wjonassen: Been a while, but I updated with an undefined check.  
Since the PR was already approved, I will just resolve and close the PR.

Issue #78, opened 2023-12-13
Use swagger annotations instead of manual edits to swagger.json for OpenAPI documentation


adamkorynta:
**Component to be refactored**
Refactor the manually edited OpenAPI swagger config https://github.com/opendcs/rest_api/blob/main/opendcs-rest-api/src/main/resources/swaggerui/swagger.json to use Java annotations instead.

**Purpose of refactoring**
Manual editing of JSON file decreases maintainability and is decoupled from the REST API Java classes where the endpoints and parameters are configured.

**Proposed method**
Use the swagger Java annotations: https://github.com/swagger-api/swagger-core/wiki/Annotations

**Additional context**
CWMS Data API uses Javalin's Swagger annotations: https://github.com/javalin/javalin-openapi

adamkorynta:
The "servers" block also needs to figure out the parameterized context automatically. I have no investigated what mechanism can do this.

wjonassen:
When working on this, we need to keep this conversation in mind involving SSO and how the rest api knows where the login URL is - 
https://github.com/opendcs/rest_api/pull/197/files#r1781997932

MikeNeilson:
Looks like we should be able to do it the same way. https://github.com/swagger-api/swagger-core/wiki/Swagger-Core-Jersey-2.X-Project-Setup-1.5

I'm pretty that's what Javalin is using under the hood.

PR #197, opened 2024-09-28
Remove backend gateway and have javascript talk directly with the API

wjonassen:
Currently, javascript hits its own webserver and then that is relayed to the api.  This is unnecessarily complicated, so now the browser will talk directly with the API, removing that middle step.

## Solution

The goal is to remove the gateway code completely and have the user set the path to the api in a config file.

## how you tested the change

Need to test that all of the pages and ajax calls interact the same as before.  NOTE - CORS settings will need to be set properly as well.


opendcs-web-client/src/main/java/portal/decodes/Schedule.java
wjonassen, line 1: Sorry,  I was doing some quick bug fixes and this file snuck in there.

opendcs-web-client/src/main/webapp/resources/js/datatables/datatables.js
wjonassen, line 1: Sorry,  I was doing some quick bug fixes and this file snuck in there.

wjonassen:
This is just a small change to begin with so I can get some comments before I make the changes for the entire project.  I've done login, as well as the platforms page.  Once we agree, the rest of the pages will be updated to follow the same process.

opendcs-web-client/src/main/webapp/WEB-INF/web.xml
```diff
@@ -27,6 +27,10 @@
         <welcome-file>default.htm</welcome-file>
         <welcome-file>default.jsp</welcome-file>
     </welcome-file-list>
+    <context-param>
+        <param-name>api_base_url</param-name>
+        <param-value>http://opendcsapi:8080/odcsapi</param-value>
```
MikeNeilson, line 32: For the moment I think it's reasonable to assume same host and "odcsapi" as the context. We can address the possible flexibility later. This is forcing HTTP instead of HTTPS and a specific domain name. 

Thus will make testing by others somewhat difficult in several situations.

opendcs-web-client/src/main/webapp/resources/jsp/includes/decodes.jsp
```diff
@@ -64,6 +63,13 @@
 <script src="../resources/js/lib/opendcs_utilities.js"></script>
 <%-- /opendcs web files --%>
 
+<%-- Get API Path --%>
+<script>
+	window.API_BASE_URL = '<%= getServletConfig().getServletContext().getInitParameter("api_base_url") %>';
```
MikeNeilson, line 68: Based on above comment, basically keep this logic but make it a clientside grab of the existing window.location and change the base path.

opendcs-web-client/src/main/webapp/resources/js/configs.js
```diff
@@ -1391,13 +1368,11 @@ document.addEventListener('DOMContentLoaded', function() {
                 var token = sessionStorage.getItem("token");
```
adamkorynta, line 1368: can remove the calls to get the token

adamkorynta (approved):

opendcs-web-client/src/main/webapp/WEB-INF/web.xml
```diff
@@ -28,8 +28,16 @@
...
-        <param-value>http://opendcsapi:8080/odcsapi</param-value>
+        <param-name>api_base_path</param-name>
+        <param-value>odcsapi</param-value>
+    </context-param>
+    <context-param>
+        <param-name>authentication_type</param-name>
+        <param-value>basic</param-value>
+    </context-param>
+    <context-param>
+        <param-name>authentication_base_path</param-name>
```
MikeNeilson, line 39: This really shouldn't need to be configured, with the changes @adamkorynta just merged in there should be sufficient feedback from the API itself to properly redirect, or list the available choices to login.

opendcs-web-client/src/main/webapp/resources/js/datatables/helpers.js
```diff
@@ -103,11 +103,9 @@ function createActionDropdown(applicableActions)
 function setPropSpecsMeta()
 {
     //seasonMeta
-    var params = { 
-            "opendcs_api_call": "seasons"
-    };
+    var params = {};
     $.ajax({
-        url: `../api/gateway`,
+        url: `${window.API_URL}/seasons`,
```
MikeNeilson, line 108: We may want to wrap these within a larger API. I know the cookies will get get sent, but I think the authorization header may not always be sent.

opendcs-web-client/src/main/webapp/WEB-INF/web.xml
adamkorynta: The REST API doesn't return a login URL (or list of URL's). It doesn't seem right that the REST API would know about the `CWMSLogin/login` endpoint, it just needs to know that the JSESSIONIDSSO was setup with an appropriate principal. If the web client, for example, were to run queries against both OpenDCS REST API and CDA, the CWMSLogin endpoint would be the same authentication for both. Would we want the client to interrogate both API's and if they support the same auth type and login URL, then only make a single call?

MikeNeilson: That's were things get a bit.. obnoxious? for us. The whole cookie thing (JSESSIONIDSSO and some of these context parameter) are really geared towards everything being from the same server.

Unfortunately for the original designers of that specification things like external identity providers (OpenID Connect) became a thing.

I suspect the OpenAPI specification is the reasonable balance. Avoids us having to design a custom JSON and write the endpoints but still provides the requisite information about the various options.

At the same time I'm not super happy with some of the limits I've felt in the spec, at least as to what the swagger will render. The actual Authorization header spec is relatively free form and specifically asserts there should be flexibility but the html/js will only render the defined standards correctly.

opendcs-web-client/src/test/java/selenium/LoginTest.java
wjonassen, line 1: This works great.  It is mostly a demonstration to show we have it set up and that we can do this on multiple browsers.  Next step is to do it as a JUnit parameterized test.

opendcs-web-client/src/main/webapp/WEB-INF/web.xml
wjonassen: Sounds good - I've addressed this by going back to base path and assuming same host.

opendcs-web-client/src/main/webapp/resources/jsp/includes/decodes.jsp
wjonassen: Sounds good - updated.

opendcs-web-client/src/test/java/selenium/LoginTest.java
```diff
@@ -0,0 +1,105 @@
...
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.firefox.FirefoxDriver;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+//@Disabled("For demonstration purposes currently.")
```
adamkorynta, line 16: Should be disabled

opendcs-web-client/src/main/webapp/WEB-INF/web.xml
wjonassen: Sounds good - 

I've referenced this conversation in the issue for swagger annotations:

https://github.com/opendcs/rest_api/issues/78

opendcs-web-client/src/test/java/selenium/LoginTest.java
wjonassen: Sorry, I meant to have that disabled.  Done.

adamkorynta (approved):

PR #207, opened 2024-10-02
Add coverage badge

MikeNeilson:
I was looking at code coverage setups and realized there's no badge specific to the current coverage.

NOTE: I will be adding this to the core code after the gradle migration. I believe it's a similar number.


adamkorynta:
Good call, I think I added the badge before there was any coverage.

adamkorynta (approved):

PR #213, opened 2024-10-25
example concept of using opendcs DAO's as a replacement for rest_api dao

adamkorynta:
this will gain access to other database implementations and reduce SQL query duplications. still needs unit tests
still needs integration tests
still needs to use artifacts produced after merging https://github.com/opendcs/opendcs/pull/681

Using this PR as the basis for a revised work plan.

## Problem Description

Need to be able to support multiple database implementations while reducing redundant SQL queries.

## Solution

Utilize work done in https://github.com/opendcs/opendcs/pull/681 to pull in the DAI interfaces from opendcs itself.

## how you tested the change

TODO

## Where the following done:


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/dao/ApiDaoBase.java
```diff
@@ -15,46 +15,64 @@
...
 
-public class ApiDaoBase
-	implements ApiDaiBase
+import decodes.sql.SqlDatabaseIO;
+import org.opendcs.odcsapi.hydrojson.DbInterface;
+import org.opendcs.odcsapi.util.ApiConstants;
+
+public class ApiDaoBase implements ApiDaiBase
 {
+	private final SqlDatabaseIO dbio;
```
MikeNeilson, line 31: This one should also be the OpenDcsDatabase class. It'll have the getDao method and an accessor to the "Database" instance (which has the DbIo if you need it.)

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/dao/DbException.java
```diff
@@ -18,28 +18,28 @@
...
 {
 	private static final long serialVersionUID = 4552935582551955795L;
-	private String module = null;
-	private Exception cause = null;
-	
-	public DbException(String msg)
-	{
-		super(msg);
-	}
+	private final String module;
```
MikeNeilson, line 21: If this is because of the existing things go ahead and just ditch it. We was used for identifying things in the logs but we're moving to slf4j with the full class names over time.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/dao/DbException.java
```diff
@@ -18,28 +18,28 @@
...
 	public DbException(String module, Exception cause, String msg)
 	{
-		this(msg);
+		super(msg, cause);
 		this.module = module;
-		this.cause = cause;
 	}
 
-	public String getModule()
+	public DbException(Exception cause, String msg)
```
MikeNeilson, line 29: 1. This just looks weird being backwards.
2. cause should be `Throwable`

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OpenDcsResource.java
```diff
@@ -0,0 +1,47 @@
...
+class OpenDcsResource
+{
+	private final OpenDcsDatabase db;
+
+	protected OpenDcsResource()
+	{
+		try
+		{
+			//TODO - need to iron out the correct way to handle data source.
+			Context initialCtx = new InitialContext();
```
MikeNeilson, line 32: https://github.com/USACE/cwms-data-api/blob/5fd3d7e6d0ead40992bc47631916245575b4d017/cwms-data-api/src/main/java/cwms/cda/ApiServlet.java#L262

This is a "servlet" thing, not a Tomcat/javalin thing. it should generally be injected on each request. Using a base class maybe still work but the DataSource becomes a class property and the getDb would actually create the database.

Debatable if that should, or even can, be a static shared variable. OpenDcsDatabase *should* be thread safe.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/dao/ApiDaoBase.java
adamkorynta: oh yeah, I need to revert changes in this PR and just deprecate the whole class

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/dao/DbException.java
adamkorynta: agreed, but the other constructor uses the first String param for the "module" so I got stuck. I think this should just be a different exception class, especially since I think all DbExceptions are getting reported at error code 400 which doesn't seem right, at least across the board.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/dao/DbException.java
adamkorynta: yeah, just too much code to update in this PR to completely ditch it, that's why I endeded up overloading. Better to just cut over to a new class though.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OpenDcsResource.java
adamkorynta: yeah that's where I was heading. Wasn't sure if we should inject the OpenDcsDatabase as a singleton/shared variable if it's thread safe which would mean we don't want the DataSource itself injected. For the moment I put OpenDcsDatabase in the constructor rather than the getter because it isn't the Closeable instance and so the lifecycle doesn't need to be managed.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/SecurityHeadersFilter.java
```diff
@@ -25,7 +25,7 @@
 
 
 @Provider
-@Priority(Priorities.AUTHENTICATION)
+@Priority(Priorities.USER)
```
MikeNeilson, line 28: Doesn't authentication have a higher priority than user?

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/TomcatServer.java
```diff
@@ -76,6 +79,9 @@ public TomcatServer(String baseDir, int port, String contextName) throws IOExcep
 		Pipeline pipeline = host.getPipeline();
 		pipeline.addValve(ssoValve);
 		context = (StandardContext) tomcatInstance.addWebapp(contextName, Paths.get("src/main/webapp").toAbsolutePath().toString());
+		context.addApplicationListener(DataSourceContextCreator.class.getName());
```
MikeNeilson, line 82: Why the change to listener vs provider?

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/opendcs_dep/TestDecoder.java
```diff
@@ -339,5 +343,33 @@ private static void decodedMsg2api(ApiDecodedMessage adm, DecodedMessage dm, Arr
...
+	private static synchronized void initDecodes(OpenDcsDatabase db)
+			throws DbException
+	{
+		// Don't init more than once per minute.
+		long now = System.currentTimeMillis();
+		if (now - lastDecodesInitMsec < 60*1000L)
+			return;
+		lastDecodesInitMsec = now;
+
+		if (decodes.db.Database.getDb() == null)
```
MikeNeilson, line 359: FYI Instances of OpenDcsDatabase are required to have a valid "Database" instance if still required, but our main intent is to move downstream code to the getDao feature and hide the Database detail, if still needed, in the getDao implementation.

adamkorynta:
created https://github.com/opendcs/rest_api/pull/271 to replace this example PR

PR #276, opened 2024-12-18
Issue #232: Refactored SiteResources Controller to use OpenDCS DAIs

### LINKED ISSUE #232: Refactor SiteResources controller

zack-rma:
## Problem Description

Fixes #232. 

## Solution

## how you tested the change

Unit testing included.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/SiteResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/SiteResources.java
@@ -18,6 +18,7 @@ package org.opendcs.odcsapi.res;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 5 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/SiteResources.java
```diff
@@ -30,99 +36,233 @@
...
+		returnSite.setCountry(site.country);
+		returnSite.setState(site.state);
+		returnSite.setProperties(properties);
+		returnSite.setNearestCity(site.nearestCity);
+		returnSite.setLatitude(site.latitude);
+		returnSite.setLongitude(site.longitude);
+		returnSite.setTimezone(site.timeZoneAbbr);
+		returnSite.setRegion(site.region);
+		returnSite.setPublicName(site.getPublicName());
+		HashMap<String, String> siteNames = new HashMap<>();
```
rma-rripken, line 169: I'd think this would be Map<String, String> and the impl class LinkedHashMap

zack-rma: I updated the impl class, but the ApiSiteRef DTO requires a HashMap object containing the sitenames.

rma-rripken: Is there a reason ApiSiteRef requires HashMap?  Thats an implementation, methods should require and return interfaces anywhere thats possible.

zack-rma: This should be able to be rectified by the recent PR to refactor the API DTOs in use: https://github.com/opendcs/rest_api/pull/317

rma-rripken (approved):

MikeNeilson:
followup PR has same commits.

PR #278, opened 2024-12-18
Issue #223: Implemented DataSourceResources Controller refactor

### LINKED ISSUE #223: Refactor DataSourceResources controller

zack-rma:
## Problem Description

Fixes #223. 

## Solution

Replaced DAO usage with OpenDCS DAI implementations.

## how you tested the change

Unit tests included.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/DataSourceResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/DataSourceResources.java
@@ -15,10 +15,12 @@
 
 package org.opendcs.odcsapi.res;
 
-import java.sql.SQLException;
-import java.util.logging.Logger;
+import java.util.ArrayList;
+import java.util.Properties;
+import java.util.Vector;
 
 import javax.annotation.security.RolesAllowed;
+import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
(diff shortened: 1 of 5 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/DataSourceResources.java
```diff
@@ -52,78 +57,231 @@ public class DataSourceResources
...
+			ads.setDataSourceId(DbKey.NullKey.getValue());
+		}
+		ads.setName(ds.getName());
+		ads.setType(ds.dataSourceType);
+		ads.setProps(ds.getArguments());
+		ads.setGroupMembers(map(ds.groupMembers));
+		return ads;
+	}
+
+	static ArrayList<ApiDataSourceGroupMember> map(Vector<DataSource> groupMembers)
```
rma-rripken, line 162: @MikeNeilson is there a reason the DTO classes have to use ArrayList and Vector?  Can we just fix the DTO so that we don't have to return implementation classes?

rma-rripken (approved):

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/DataSourceResources.java
MikeNeilson: Not really. In a few cases it's because the Original DAOs or Interfaces used Vector and they ArrayList. So there would need to be a conversion step but I would argue we should just do so.

The DAOs are going to be moved to List or Collection as the output, but we haven't gotten to it yet.

zack-rma:
Closed since new IT PR contains this code already: https://github.com/opendcs/rest_api/pull/323

PR #283, opened 2024-12-19
Refactor PlatformResources Controller

### LINKED ISSUE #228: Refactor PlatformResources controller

zack-rma:
## Problem Description

Fixes #228. 

## Solution

## how you tested the change

Unit tests of data mapping included.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
@@ -16,8 +16,10 @@
 package org.opendcs.odcsapi.res;
 
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.Vector;
 
@@ -68,14 +70,14 @@ public class PlatformResources extends OpenDcsResource
 	public Response getPlatformRefs(@QueryParam("tmtype") String tmtype)
 			throws DbException
 	{
-		HashMap<String, ApiPlatformRef> ret = new HashMap<>();
+		Map<String, ApiPlatformRef> ret = new LinkedHashMap<>();
 		try
 		{
 			// TODO: Add support for filtering by transport media type
 			DatabaseIO dbio = getLegacyDatabase();
 			PlatformList platformList = new PlatformList();
 			dbio.readPlatformList(platformList);
-			ArrayList<ApiPlatformRef> platSpecs = map(platformList);
+			List<ApiPlatformRef> platSpecs = map(platformList);
 			for(ApiPlatformRef ps : platSpecs)
 				ret.put(ps.getName(), ps);
 			return Response.status(HttpServletResponse.SC_OK).entity(ret).build();
@@ -86,9 +88,9 @@ public class PlatformResources extends OpenDcsResource
 		}
 	}
 
-	static ArrayList<ApiPlatformRef> map(PlatformList platformList)
+	static List<ApiPlatformRef> map(PlatformList platformList)
 	{
-		ArrayList<ApiPlatformRef> ret = new ArrayList<>();
+		List<ApiPlatformRef> ret = new ArrayList<>();
 		Iterator<Platform> platform = platformList.iterator();
 		while (platform.hasNext())
 		{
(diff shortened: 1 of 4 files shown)


zack-rma:
References getAll method may require new OpenDCS DAI method to support filtering by transport media type

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
```diff
@@ -33,113 +35,380 @@
...
+
+	static ArrayList<ApiPlatformRef> map(PlatformList platformList)
+	{
+		ArrayList<ApiPlatformRef> ret = new ArrayList<>();
+		Iterator<Platform> platform = platformList.iterator();
+		while (platform.hasNext())
+		{
+			ApiPlatformRef ref = new ApiPlatformRef();
+			Platform plat = platform.next();
+			ref.setName(plat.getDisplayName());
```
rma-rripken, line 97: Can this call the single object map() method?

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
```diff
@@ -33,113 +35,380 @@
...
 	@GET
 	@Path("platformrefs")
 	@Produces(MediaType.APPLICATION_JSON)
 	@RolesAllowed({AuthorizationCheck.ODCS_API_GUEST})
-	public Response gePlatformRefs(@QueryParam("tmtype") String tmtype)
-		throws DbException, SQLException
+	public Response getPlatformRefs(@QueryParam("tmtype") String tmtype)
+			throws DbException
 	{
 		HashMap<String, ApiPlatformRef> ret = new HashMap<>();
```
rma-rripken, line 71: HashMap vs Map ?  Is this driven by DTO?   I'm always suspicious of HashMap.  IMO the extra cost of LinkedHashMap is always worth it to maintain insertion order, even if its just to make debugging consistent.

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/PlatformResourcesTest.java
```diff
@@ -0,0 +1,427 @@
...
+		assertFalse(apiStatus.isEmpty());
+		assertEquals(1, apiStatus.size());
+		ApiPlatformStatus apiStat = apiStatus.get(0);
+		assertEquals(status.getAnnotation(), apiStat.getAnnotation());
+		assertEquals(status.getLastContactTime(), apiStat.getLastContact());
+		assertEquals(status.getPlatformId().getValue(), apiStat.getPlatformId());
+		assertEquals(status.getLastRoutingSpecName(), apiStat.getRoutingSpecName());
+	}
+
+	private static int iterSize(Iterator<?> it)
```
rma-rripken, line 417: interesting.  
+1 b/c its clear what this is doing but -1 b/c its weird to have to do this.

rma-rripken (approved):

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
zack-rma: Since the ApiPlatformRef has a different structure to ApiPlatform, the map can't be reused.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
zack-rma: I've swapped HashMap for Map and the implementation for LinkedHashMap. I've also replaced ArraList with List where possible/appropriate.

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/PlatformResourcesTest.java
zack-rma: Yeah, if the implementation gave us more than just the iterator this wouldn't be needed.

MikeNeilson: These will be migrated to a more standard DAO style, but platforms is not first on the list (Enum is in the works, routingspecs/schedule entries are probably next) so for now you're stuck with doing this.

zack-rma:
Closing due to newer PR with same changes: https://github.com/opendcs/rest_api/pull/312

PR #285, opened 2024-12-19
PresentationResources Controller Refactor with Integration Testing for OpenTSDB

### LINKED ISSUE #229: Refactor PresentationResources controller

### LINKED ISSUE #241: OpenTSDB: Create Integration Tests for PresentationResources controller

zack-rma:
## Problem Description

Fixes #229. 
Fixes #241. 

## Solution

## how you tested the change

Unit tests included.

## Where the following done:

   - [x] Integration tests created or modified that run during integration testing

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
@@ -65,10 +65,9 @@ public class PresentationResources extends OpenDcsResource
 	@RolesAllowed({ApiConstants.ODCS_API_GUEST})
  	public Response getPresentationRefs() throws DbException
 	{
-		DatabaseIO dbIo = null;
+		DatabaseIO dbIo = getLegacyDatabase();
 		try
 		{
-			dbIo = getLegacyDatabase();
 			PresentationGroupList groupList = new PresentationGroupList();
 			dbIo.readPresentationGroupList(groupList);
 			return Response.status(HttpServletResponse.SC_OK).entity(map(groupList)).build();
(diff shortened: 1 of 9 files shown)


zack-rma:
Will need new OpenDCS DAI method implementation for full feature support

zack-rma:
Requires Toolkit method implementations found here: https://github.com/adamkorynta/opendcs/pull/8

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
```diff
@@ -31,95 +33,294 @@
...
+		else
+		{
+			presGrp.setGroupId(DbKey.NullKey.getValue());
+		}
+		presGrp.setElements(map(group.dataPresentations));
+		return presGrp;
+	}
+
+	static ArrayList<ApiPresentationElement> map(Vector<DataPresentation> dataPresentations)
+	{
```
rma-rripken, line 178: Vector?  You should be able to use more general interface types like List in the method arguments -right?   I understand you are returning ArrayList b/c you need to pass the result of the call into something that requires ArrayList but shouldn't your methods be able to accept more general arguments?

rma-rripken (approved):

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
zack-rma: I was able to generalize this mapping, it now works with lists.

rma-rripken (approved):

rma-rripken (approved):

PR #287, opened 2024-12-19
Refactor OdcsapiResource Controller

### LINKED ISSUE #227: Refactor OdcsapiResource controller

zack-rma:
## Problem Description

Fixes #227. 

## Solution

## how you tested the change

Unit tests for data mapping included.

## Were the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
@@ -18,6 +18,7 @@ package org.opendcs.odcsapi.res;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.Properties;
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletResponse;
@@ -45,14 +46,12 @@ import decodes.decoder.TimeSeries;
 import decodes.sql.DbKey;
 import decodes.tsdb.DbIoException;
 import decodes.tsdb.TimeSeriesDb;
-import decodes.util.PropertySpec;
 import ilex.var.NoConversionException;
 import ilex.var.TimedVariable;
 import ilex.var.Variable;
 import org.opendcs.database.api.OpenDcsDatabase;
 import org.opendcs.odcsapi.beans.ApiDecodedMessage;
 import org.opendcs.odcsapi.beans.ApiDecodesTimeSeries;
-import org.opendcs.odcsapi.beans.ApiPropSpec;
 import org.opendcs.odcsapi.beans.ApiRawMessage;
 import org.opendcs.odcsapi.beans.DecodeRequest;
 import org.opendcs.odcsapi.dao.DbException;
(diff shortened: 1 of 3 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
```diff
@@ -54,11 +73,23 @@ public class OdcsapiResource extends OpenDcsResource
...
 		{
-			return ApiHttpUtil.createResponse(dao.getTsdbProperties());
+			TimeSeriesDb tsdb = getLegacyTimeseriesDB();
+			tsdb.readTsdbProperties(tsdb.getConnection());
+
+			Properties props = new Properties();
+			for (Enumeration<String> propertiesEnumeration = tsdb.getPropertyNames();
+				 propertiesEnumeration.hasMoreElements();)
+			{
+				String key = propertiesEnumeration.nextElement();
```
rma-rripken, line 85: Does enhanced-for not work here?  I don't see an error but it seems like it could be simplier.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
```diff
@@ -54,11 +73,23 @@ public class OdcsapiResource extends OpenDcsResource
...
+				props.setProperty(key, tsdb.getProperty(key));
+			}
+			return Response.status(HttpServletResponse.SC_OK).entity(props).build();
+		}
+		catch(SQLException e)
+		{
+			throw new DbException("Error reading timeseries properties: " + e);
```
rma-rripken, line 92: This just builds new exception with the toString of the old exception.  Is there a ctor that lets you wrap the sql exception?

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
```diff
@@ -101,9 +147,87 @@ public Response getPropSpecs(@QueryParam("class") String className)
...
+		try
+		{
+			DecodedMessage message = map(TestDecoder.decodeMessage(request.getRawmsg(),
+					request.getConfig(), scriptName, db), request.getRawmsg());
+			return Response.status(HttpServletResponse.SC_OK).entity(message).build();
+		}
+		catch (UnknownPlatformException | DatabaseException e)
+		{
+			throw new DbException("Unknown platform: " + e.getMessage());
+		}
```
rma-rripken, line 164: Wondering about wrapping the exception

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
zack-rma: Swapped for an enhanced for. I had to wrap the Enumeration, since it doesn't support enhanced-for by itself.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
zack-rma: I swapped this usage for one that takes the cause as a parameter

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
zack-rma: Swapped for usage with cause as a parameter.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OdcsapiResource.java
rma-rripken: I don't use Enumeration often, if it doesn't support it -fine. Or, sounds like you already changed it so leave it.  The hasMoreElements/nextElement stood out

zack-rma:
Closed since IT PR has this code and the unit tests are no longer relevant (no data mapping in controller): https://github.com/opendcs/rest_api/pull/306

PR #274, opened 2024-12-17
Issue #219: Refactor AppResources Controller

zack-rma:
## Problem Description

Fixes #219 . 

## Solution

Replaces DAO usage with OpenDCS DAI implementations.

## how you tested the change

Unit tests included.

## Where the following done:

- [ ] Tests. Check all that apply:
   - [x] Unit tests created or modified that run during ant test.
   - [ ] Integration tests created or modified that run during integration testing
         (Formerly called regression tests.)
   - [ ] Test procedure descriptions for manual testing
- [ ] Was relevant documentation updated?
- [ ] Were relevant config element (e.g. XML data) updated as appropriate

If you aren't sure leave unchecked and we will help guide you to want needs changing where.

PR #311, opened 2024-12-30
AppResources Controller Refactor with Integration Testing for OpenTSDB

### LINKED ISSUE #220: OpenTSDB: Create Integration Tests for AppResources controller

zack-rma:
## Problem Description

Fixes #220. 

## Solution
Implements integration testing against OpenTSDB.

Builds upon changes in #274 

## how you tested the change
Includes integration tests. 

## Where the following done:

   - [x] Integration tests created or modified that run during integration testing


zack-rma:
Currently there exists a failing test case due to an unknown `EventPort` value for the `compproc` app used in testing. This is causing a connection failure, with the specific location of the issue identified by `TODO` tasks.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -35,67 +37,97 @@
...
+					.entity(mapLoading(dai.getComputationApp(DbKey.createDbKey(appId)))).build();
+		}
+		catch (NoSuchObjectException e)
+		{
+			return Response.status(HttpServletResponse.SC_NOT_FOUND)
+					.entity(String.format(NO_APP_FOUND, appId)).build();
+		}
+		catch (DbIoException ex)
+		{
+			throw new DbException("No such app with ID " + appId, ex);
```
rma-rripken, line 130: Isn't this just your new constant?

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -105,112 +137,202 @@ public Response getApp(@QueryParam("appid") Long appId)
...
+
 				Integer port = appStat.getEventPort();
 				if (port == null)
-					return ApiHttpUtil.createResponse(new ArrayList<ApiAppEvent>()); // app not running
+				{
+					return Response.status(HttpServletResponse.SC_OK)
+							.entity(new ArrayList<ApiAppEvent>()).build();
+				}
 				apiEventClient = new ApiEventClient(appId, appStat.getHostname(), port, appStat.getAppName(), appStat.getPid());
 				apiEventClient.connect();
```
rma-rripken, line 324: This logic is being duplicated.  Its unclear to me why the first branch doesn't need to check appStat.getPid() != null but the second branch does.  In the second branch you don't need to check cli.ifPresent it has to be present or you would be in the first branch.  I think these two could be combined and the code would be clearer.

MikeNeilson:
Not sure if it affects much of the current work but fyi we are planning on
removing that event system in favor of standard monitoring tools, likely
opentelemtry.

Which is not to say you dont have to deal with it just maybe dont put
excessive energy in beyond making sure things start. Might be easier to
make sure it's more optional at the OpenDCS core level.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -105,112 +137,202 @@ public Response getApp(@QueryParam("appid") Long appId)
...
-			return ApiHttpUtil.createResponse(apiEventClient.getNewEvents());
+			return Response.status(HttpServletResponse.SC_OK)
+					.entity(apiEventClient.getNewEvents()).build();
 		}
 		catch(ConnectException ex)
-		{			
+		{
 			throw new WebAppException(ErrorCodes.IO_ERROR,
 					String.format("Cannot connect to %s.", appStat.getAppName()), ex);
 			// NOTE: event client added to user token ONLY if connect succeeds.
```
rma-rripken, line 338: I think this is trying to point out that if the "connect()" call throws an exception then the clientConnectionCache doesn't get anything added to it.  This comment is weirdly worded for where the comment is located.  A comment about what happens when the connect succeeds should probably be at the connect call or somewhere on the happy path.  At this location the comment should be reworded so its clear what didn't happen.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -222,86 +344,166 @@ else if (appStat.getPid() != null && appStat.getPid() != cli.get().getPid())
...
 			// ProcWaiterThread runBackground to execute command, use callback.
 			ProcWaiterCallback pwcb = (procName, obj, exitStatus) ->
-					{
-						ApiLoadingApp loadingApp1 = (ApiLoadingApp)obj;
-						LOGGER.info("App Termination: app {} was terminated with exit status {}",
-								loadingApp1.getAppName(), exitStatus);
-					};
+			{
+				ApiLoadingApp loadingApp1 = (ApiLoadingApp)obj;
+			};
```
rma-rripken, line 384: I don't understand what this accomplishes.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
```diff
@@ -222,86 +344,166 @@ else if (appStat.getPid() != null && appStat.getPid() != cli.get().getPid())
...
+					return map(dai, lock);
+				}
+			}
+			List<ApiAppRef> apps = dai.listComputationApps(false)
+					.stream()
+					.map(AppResources::map)
+					.collect(Collectors.toList());
+			for (ApiAppRef app : apps)
+			{
+				if (app.getAppId().equals(appId))
```
rma-rripken, line 492: Can't this be combined with the streaming call on 488?

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OpenDcsResource.java
```diff
@@ -81,7 +84,19 @@ final OpenDcsDatabase createDb()
 			{
 				throw new IllegalStateException("Error connecting to the database via JNDI", ex);
 			}
-//			throw new IllegalStateException("Error connecting to the database via JNDI", e);
+			//			throw new IllegalStateException("Error connecting to the database via JNDI", e);
```
rma-rripken, line 87: If its handled, just delete this commented out throw call.

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/AppResourcesTest.java
```diff
@@ -0,0 +1,117 @@
...
+		compAppInfo.setProperties(properties);
+
+		ApiAppRef appRef = map(compAppInfo);
+
+		assertNotNull(appRef);
+		assertEquals(compAppInfo.getAppName(), appRef.getAppName());
+		assertEquals(compAppInfo.getComment(), appRef.getComment());
+		assertEquals(compAppInfo.getLastModified(), appRef.getLastModified());
+		assertEquals(compAppInfo.getAppId().getValue(), appRef.getAppId());
+		assertEquals(compAppInfo.getAppType(), appRef.getAppType());
```
rma-rripken, line 42: Somewhere in this method should be a comment about Properties.  ApiAppRef doesn't have a place to hold the properties -right?   Its good to show that map() isn't broken by presence/absence of Properties on the input but I think it would be nice to have a comment pointing out that they aren't mapped.

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/AppResourcesTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+	@Test
+	void testLoadingAppMap()
+	{
+		CompAppInfo compAppInfo = new CompAppInfo();
+		compAppInfo.setAppName("Dijkstra's Algorithm");
+		compAppInfo.setComment("Runs Dijkstra's algorithm on a graph");
+		compAppInfo.setLastModified(Date.from(Instant.parse("2021-07-01T00:00:00Z")));
+		compAppInfo.setAppId(DbKey.createDbKey(151615L));
+		compAppInfo.setNumComputations(1);
```
rma-rripken, line 101: This doesn't get mapped - right? Add a comment that it doesn't

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/AppResourcesIT.java
```diff
@@ -0,0 +1,474 @@
...
+		for (int i = 0; i < actual.getList("").size(); i++)
+		{
+			String actualAppName = actual.get("[" + i + "].appName");
+			String expectedAppName = expected.get("[" + i + "].appName");
+			if (actualAppName.equalsIgnoreCase(expectedAppName))
+			{
+				assertEquals(actualAppName, expectedAppName);
+				assertEquals(actual.get("[" + i + "].appType").toString(),
+						expected.get("[" + i + "].appType").toString());
+				assertEquals(actual.get("[" + i + "].comment").toString(),
```
rma-rripken, line 118: All this String manipulation isn't very elegant.  Can't you hang on to the list from the getList("") call and do normal List stuff with that?

rma-rripken (changes requested):
Resolve my other comments

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/AppResourcesIT.java
zack-rma: Cleaned up to remove String manipulation

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
zack-rma: Replaced with constant

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
zack-rma: Cleaned up duplicated logic

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
zack-rma: I moved the comment to a more appropriate location and added a comment on why the exception may be thrown

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
zack-rma: Looks like this was just used for logging when an application fails. I removed the usage

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OpenDcsResource.java
zack-rma: Removed

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/AppResourcesTest.java
zack-rma: Added a properties comment and notes for other data lost in mapping

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java
zack-rma: Cleaned up the logic here

rma-rripken (approved):

rma-rripken (approved):

zack-rma:
After some discussion, it seems this controller will need to have its functionality verified. It is unclear if this was used in the web UI and whether it was operational before the refactor. There is no current method of obtaining a process ID when starting an app, which is required to obtain a lock in the `CP_COMP_PROC_LOCK` table. This table is also used to check for running processes and their statuses. We'll likely need to investigate the Toolkit's methods of obtaining a PID and see if it can be used here.

adamkorynta:
superceded by: https://github.com/opendcs/rest_api/pull/354

PR #326, opened 2025-01-23
Feature/cwms db implementation tests

### LINKED ISSUE #245: Add CWMS implementation to integration test harness
- Needs to utilize CWMS/CCP schema running in a docker container
- Add authorization mechanism for integration testing

adamkorynta:
## Problem Description

Fixes #245. 
Need to support integration testing against CWMS CCP schema.

## Solution

Use the updates OpenDCS Toolkit fixtures that support installing CCP via flyway.

## how you tested the change

Automated testing of the AlgorithmResources endpoint.

## Where the following done:


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/AuthorizationCheck.java
```diff
@@ -15,23 +15,103 @@
...
+		{
+			//Temporary workaround until database_properties table is implemented in the schema
+			LOGGER.atWarn().setCause(e).log("Temporary solution forcing OpenTSDB");
+			DecodesSettings decodesSettings = new DecodesSettings();
+			decodesSettings.CwmsOfficeId = System.getProperty("DB_OFFICE");
+			try(Connection connection = dataSource.getConnection())
+			{
+				DatabaseProvider databaseProvider;
+				String databaseProductName = connection.getMetaData().getDatabaseProductName();
+				if(databaseProductName.toLowerCase().startsWith("oracle"))
```
MikeNeilson, line 101: fyi there is an opendcs-oracle implementation.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/AuthorizationCheck.java
```diff
@@ -15,23 +15,103 @@
...
+				{
+					databaseProvider = new CwmsDatabaseProvider();
+				}
+				else
+				{
+					databaseProvider = new OpenTsdbProvider();
+				}
+				return databaseProvider.createDatabase(dataSource, decodesSettings);
```
MikeNeilson, line 109: any particular reason the DatabaseService factory doesn't work? i suspect the decodessettings + office issue but want to confirm.

opendcs-rest-api/build.gradle
```diff
@@ -166,10 +169,36 @@ tasks.register('integrationTest', Test) {
...
+    shouldRunAfter test
+    dependsOn war
+    dependsOn generateConfig
+    dependsOn downloadAndUnzipInstaller
+    jvmArgs += "-DwarContext=/opendcs-rest-api"
+    jvmArgs += "-Djava.util.logging.config.file=$projectDir/test/resources/logging.properties"
+    jvmArgs += "-Dorg.apache.tomcat.util.digester.PROPERTY_SOURCE=org.apache.tomcat.util.digester.EnvironmentPropertySource"
+    jvmArgs += "-Dcatalina.base=${layout.buildDirectory.file("tomcat")}"
+    jvmArgs += "-Dopendcs.test.integration.db=CWMS-Oracle"
+    jvmArgs += "-Dopendcs.cwms.oracle.image=registry-public.hecdev.net/cwms/database-ready-ora-23.5:24.12.04-RC3"
```
MikeNeilson, line 198: the non-RC is available now.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/basicauth/OpenTsdbAuthorizationDAO.java
```diff
@@ -48,43 +47,44 @@ public Set<OpenDcsApiRoles> getRoles(String username) throws DbException
 		String q = "select pm.roleid, pr.rolname from pg_auth_members pm, pg_roles pr"
```
MikeNeilson, line 47: ... we really need to create a proper application level users table. /end rant.

MikeNeilson: that said definitely better than logging in as the user to check.

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/SingleSignOnWrapper.java
```diff
@@ -23,15 +23,18 @@
 import org.apache.catalina.connector.Request;
 import org.apache.catalina.connector.Response;
 
-public class SingleSignOnWrapper extends SingleSignOn{
-    public void wrappedRegister(String ssoId, Principal principal, String authType,
-    String username, String password) {
-        this.register(ssoId, principal, authType, username, password);
-    }
+public class SingleSignOnWrapper extends SingleSignOn
```
MikeNeilson, line 26: shouldn't we be able to just turn the sso valve on in the tomcat setup?

MikeNeilson (commented):
is dawned on me writing this, we can cache the created OpenDcsDatabase object, it takes and appropriately uses the datasource which woukd be valid for the life of the instance; and is responsible for maintaining the cache objects from the database.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/AuthorizationCheck.java
adamkorynta: turned into "how much code do I want to write that will get thrown away anyways"

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/SingleSignOnWrapper.java
adamkorynta: I copied this code from CDA. Isn't this bypassing the need to add a separate service that performs the auth?

MikeNeilson: Well, I don't remember that at all, but it definitely makes sense.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/AuthorizationCheck.java
MikeNeilson: Would it make sense to allow passing a DecodesSettings, or even just a `java.util.Properties` in even with the DataSource.

Like, eventually with a merge which the passed in one would override anything in the database?

adamkorynta: That's what I was considering but couldn't think of the use-case outside of this temporary solution. Merging properties from the file system, process runtime, etc with the database seems like asking for a maintenance headache.

MikeNeilson: Yeah. I can kinda see a case for it, but how often would it come up? Though the merge I had in mind was a simple (if just using java.util.Properties) passed in value of same name wins. But for now we'll leave it. Not like it's super difficult to add such support later if we decide it's required.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/OpenDcsResource.java
```diff
@@ -37,55 +38,64 @@
 
 import static org.opendcs.odcsapi.res.DataSourceContextCreator.DATA_SOURCE_ATTRIBUTE_KEY;
 
-class OpenDcsResource
+public class OpenDcsResource
```
zack-rma, line 41: Making this public should certainly make some of the integration testing smoother on the data setup side of things

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/BaseIT.java
```diff
@@ -102,28 +108,39 @@ JsonPath getJsonPathFromResource(String filename)
...
+		session.setPrincipal(mcup);
+		session.activate();
+		tomcat.getSsoValve()
+				.wrappedRegister(COOKIE, mcup, "CLIENT-CERT", null,null);
+
+
+		//Check while passing in cookie
+		given()
+			.log().ifValidationFails(LogDetail.ALL, true)
+			.accept("application/json")
```
zack-rma, line 132: There is a `MediaType.APPLICATION_JSON` variable that can be used in the accept method

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/OdcsapiResourceIT.java
```diff
@@ -28,7 +42,7 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-@Tag("integration")
+@Tag("integration-opentsdb-only")
```
zack-rma, line 45: Will all integration tags need to be updated, or will the `integration` tag run both OpenTSDB and CWMS tests?

zack-rma (commented):
Looks good, glad this is nearing the finish line

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/OdcsapiResourceIT.java
adamkorynta: `integration` runs both. ideally we'll have everything marked as `integration` when we're all done

MikeNeilson: Not that we should do it in this PR, but would it make sense to move the integration tests to it's own subproject?

Then it could consume both the server side and the GUI side (using selenium or something) elements and it's a bit more obvious what the expectation are.

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/BaseIT.java
adamkorynta: good call, updated in a couple places

zack-rma (commented):
Looks good to me

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/res/it/OdcsapiResourceIT.java
adamkorynta: yeah that's a good idea, I'll add an issue and we can look at it after the existing PR's get merged. We haven't gotten selenium testing going yet, but having it run in the same process as the RestAssured tests will simplify things

PR #333, opened 2025-02-03
Feature/setup tomcat runner

adamkorynta:
## Problem Description

Need test application to run against to demonstrate CWMS and OpenTSDB implementation functions

## Solution

Created an embedded Tomcat javaexec task that loads the REST API and Web Client wars, starts up a database (or uses an existing one), and sets up mocked authentication.

## how you tested the change

Manual testing of javaexec using OpenTSDB and CWMS implementations.

## Where the following done:

- [x] Was relevant documentation updated?


adamkorynta:
still need to write up descriptions in the readme

adamkorynta:
added developer docs

zack-rma (commented):
Looks like the build is failing, perhaps it just needs to be rerun. Otherwise, all looks good.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/SecurityHeadersFilter.java
```diff
@@ -35,8 +37,16 @@ public void filter(ContainerRequestContext request, ContainerResponseContext res
...
 		response.getHeaders().putSingle("X-Content-Type-Options", "nosniff");
 		response.getHeaders().remove("Server");
+		//swagger returns html/js/css/png/etc
+		if(request.getUriInfo()
+				.getMatchedResources()
+				.stream()
+				.map(Object::getClass)
+				.noneMatch(c -> c == SwaggerResources.class))
+		{
+			response.getHeaders().putSingle("Content-Type", "application/json");
```
zack-rma, line 49: You should be able to use a `MediaType` constant instead of the `application/json` string here

zack-rma (approved):
One small change, otherwise all looks good

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/sec/SecurityHeadersFilter.java
adamkorynta: updated

PR #336, opened 2025-02-05
update exception handling to ensure causes aren't dropped

adamkorynta:
## Problem Description

Exceptions were being lost when rethrown.

## Solution

Add the exceptions as the cause parameter.

## how you tested the change

Integration testing.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
@@ -23,7 +23,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Vector;
-
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.Consumes;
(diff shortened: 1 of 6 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
```diff
@@ -165,6 +164,10 @@ public Response getPlatform(@QueryParam("platformid") Long platformId)
 			platform = dbIo.readPlatform(platform);
 			return Response.status(HttpServletResponse.SC_OK).entity(map(platform)).build();
 		}
+		catch(ValueNotFoundException ex)
+		{
+			throw new DatabaseItemNotFoundException("Platform with ID " + platformId + " not found.");
```
zack-rma, line 169: Missing cause pass through here

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
```diff
@@ -139,6 +139,10 @@ public Response getPresentation(@QueryParam("groupid") Long groupId)
 			dbIo.readPresentationGroup(group);
 			return Response.status(HttpServletResponse.SC_OK).entity(map(group)).build();
 		}
+		catch (ValueNotFoundException e)
+		{
+			throw new DatabaseItemNotFoundException(String.format("Presentation group with ID %s not found", groupId));
```
zack-rma, line 144: Same as above

zack-rma (commented):
Not sure if you meant to leave the cause out on a couple of these

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
adamkorynta: fixed

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
adamkorynta: fixed

zack-rma (approved):

PR #338, opened 2025-02-07
Remove configuration usage for PlatformResources integration testing

zack-rma:
## Problem Description

The integration testing for PlatformResources relies on storing data via methods added to the OpenDCS Configuration class. These methods should not be in that class, so a refactor is needed.

## Solution

Swaps usage of methods in the Configuration class for DAO calls via the DAO access methods in the Configuration class.

## how you tested the change

Integration tested in the REST API against an OpenTSDB instance.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/DatabaseSetupExtension.java
+++ b/opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/DatabaseSetupExtension.java
@@ -21,9 +21,12 @@ import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import javax.servlet.http.HttpServletResponse;
 
+import decodes.db.DatabaseException;
 import decodes.db.PlatformStatus;
 import decodes.db.ScheduleEntryStatus;
 import io.restassured.RestAssured;
+import opendcs.dai.PlatformStatusDAI;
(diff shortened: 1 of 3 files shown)


opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/DatabaseSetupExtension.java
```diff
@@ -167,14 +170,28 @@ public static void loadXMLDataIntoDb(String[] files) throws Exception
 		currentConfig.loadXMLData(filePaths, new SystemExit(), new SystemProperties());
 	}
 
-	public static void storeScheduleEntryStatus(ScheduleEntryStatus status) throws Exception
+	public static void storeScheduleEntryStatus(ScheduleEntryStatus status) throws DatabaseException
```
adamkorynta, line 173: These methods don't belong here either, they should be either in a base integration test class or a utility method

adamkorynta (changes requested):

opendcs-rest-api/src/test/java/org/opendcs/odcsapi/fixtures/DatabaseSetupExtension.java
zack-rma: Moved methods to BaseIT

adamkorynta (approved):

PR #360, opened 2025-02-28
Add Swagger Specification Generation from Annotations

zack-rma:
## Problem Description

Fixes https://github.com/opendcs/rest_api/issues/78. 

## Solution

Adds Swagger API specification generation from annotations. Serves Swagger UI as a webpage.

## how you tested the change

Live Swagger UI page created when using `run` task, compared to old hard-coded specification.

## Where the following done:


opendcs-rest-api/src/main/webapp/swaggerui/swagger-initializer.js
```diff
@@ -0,0 +1,35 @@
...
+ *  limitations under the License.
+ *
+ */
+
+window.onload = function() {
+  //<editor-fold desc="Changeable Configuration Block">
+
+  // the following lines will be replaced by docker/configurator, when it runs in a docker-container
+  window.ui = SwaggerUIBundle({
+    url: "../openapi.json",
```
MikeNeilson, line 21: Is the jaxrs swagger generating this on build or providing an endpoint to query?

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java
```diff
@@ -0,0 +1,1544 @@
...
+
+public class ResourceExamples
+{
+	private ResourceExamples() {}
+
+	public static class ConfigExamples
+	{
+		private ConfigExamples() {}
+
+		public static final String BASIC = "{\n" +
```
MikeNeilson, line 26: given we sadly aren't on java, 17 wasn't it that had """ support?, anyways.

Since this are static, it might be easier on everyone to put them in resources files and this class could load the actual value. Maybe a dirty trick with reflection with the class name and field to avoid duplicating the logic, to read them in. Would avoid the excessive " and \n you're having to deal with.

... and would make it easier to deal with the multiple implementation situtation, we can follow the same "DEFAULT", "CWMS-ORACLE" ,etc naming we did with the tests.

MikeNeilson (commented):
I wonder if we can do this with CDA as well, though this is clearly a different library the lookup concept I method might work.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java
zack-rma: I gave loading from files a shot. It seems the `@ExampleObject` annotation only accepts a string constant as a value and doesn't want to cooperate with any of the loading methods I've tried. It may be possible to extend the annotation class in such a way to allow for this usage, but I'm not sure I'm familiar enough with the inner workings of Swagger Annotations to do so successfully without going down a bit of a rabbit hole.

MikeNeilson:
It wouldn't be swagger but java, but the field probably needs to be final.
So maybe it can be done in a constructor or static block. I'll try to
experiment a bit as well.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java
MikeNeilson: https://stackoverflow.com/questions/29913722/final-variable-from-resources-file

This seems to imply things can work, the static method may not, but maybe, but a `static {}` block that loops through the fields should the trick.

Totally understand if you can't get to that, and I'll try the thought later today and will open a PR to this if it works.

zack-rma: I'll take a look and give it a try

MikeNeilson: Yeah, I was thinking about that as well, actually generating the class really wouldn't be too difficult, but definitely more effort than warranted at the moment.

I'm going to dig a little two, arguably that trick above *should* count as a compile time constant, but that likely depends on exactly how swagger does things.

MikeNeilson: ... wow that is annoyingly unequivocal:

```
Compiling with toolchain '/usr/lib/jvm/java-17-openjdk-amd64'.
Compiling with JDK Java compiler API.
/home/mike/projects/rest_api/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AppResources.java:143: error: element value must be a constant expression
                                                        @ExampleObject(name = "Basic", value = ResourceExamples.AppExamples.BASIC),
```

That was just moving the actual string constant into a static block, not even trying to read a resource. there is an externalValue variant but I suspect it's not processed during that annotation processing, just passed on to the output.

MikeNeilson: yep, just a password. I guess we wait for being able to do java 17 to clean that pile of quotes up.... ugh.

doc-source/Swagger-README.md
```diff
@@ -0,0 +1,97 @@
...
+https://[REST_API_URL]/odcsapi/swaggerui/index.html
+```
+
+On development machines running the REST API locally, this will be located at 
+```
+http://localhost:7000/odcsapi/swaggerui/index.html
+```
+
+> [!NOTE]
+> The Swagger UI is only available when the REST API is running.
```
adamkorynta, line 22: Kinda silly to point this out...

doc-source/Swagger-README.md
```diff
@@ -0,0 +1,97 @@
...
+
+## OpenAPI Specification Generation
+
+In order to provide multiple options for developers to interact with the API,
+the OpenAPI specification is generated in one of two ways.
+
+### Automated Generation
+To automatically generate the OpenAPI specification, initiate the `run` Gradle task.
```
adamkorynta, line 30: Change `run` to `./gradlew run` and you'll get a nice link in IntelliJ

adamkorynta: this tip applies to the other commands below as well

doc-source/Swagger-README.md
```diff
@@ -0,0 +1,97 @@
...
+> Generated JSON and YAML OpenAPI specifications can be viewed by copying the content of the file and pasting it into the [Swagger Editor](https://editor.swagger.io/).
+
+To manually generate the OpenAPI specification, run the `generateOpenAPI` Gradle task found within
+the `documentation` group. The default output format is JSON.
+
+The manually generated OpenAPI specification will be placed in the `/build/swagger` directory. The
+file will be named `OpenDCS-REST-API.json`.
+
+To change the output format from the default JSON to YAML, edit the `generateOpenAPI` task in the
+`build.gradle` file for the opendcs-rest-api project. Change the `outputFormat` parameter from
```
adamkorynta, line 61: Developers shouldn't need to update build.gradle. I'd recommend that you either remove this from the documentation, provide two tasks, or add a gradle property.

doc-source/Swagger-README.md
```diff
@@ -0,0 +1,97 @@
...
+
+Example input data for the `POST` endpoints can be found in the
+[ResourceExamples](../opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java) 
+class, located in the `org.opendcs.odcsapi.res` package. These examples provide various levels of 
+input data for the different endpoints, which can be useful for determining the expected input data 
+for each endpoint.
+
+Adding new examples to the
+[ResourceExamples](../opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java) 
+class is simple. Create a new inner class for the endpoint resource and add a
```
adamkorynta, line 89: Avoid terms like "simple" 

Also why do you need an inner class?

opendcs-rest-api/build.gradle
```diff
@@ -255,3 +259,28 @@ tasks.register('integrationTestReportOpenTSDB', JacocoReport) {
...
+    outputDir = file('build/swagger/')
+}
+
+tasks.register('generateOpenAPI') {
+    group = "documentation"
+    description = "Generates the Swagger OpenAPI JSON file without running the application"
+    dependsOn('resolve')
+}
+
+tasks.register('cleanOpenAPI') {
```
adamkorynta, line 279: why is this needed? the `generateOpenAPI` doesn't overwrite?
regular `./gradlew clean` will clear out the whole build directory

adamkorynta (changes requested):

doc-source/Swagger-README.md
zack-rma: Removed

doc-source/Swagger-README.md
zack-rma: I've updated the task references, is a Plugin needed to see the IntelliJ link?

doc-source/Swagger-README.md
zack-rma: Added a gradle property with a default value

opendcs-rest-api/build.gradle
zack-rma: It does, I've removed the `cleanOpenAPI` task and updated the documentation accordingly

.github/workflows/publish.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         uses: gradle/gradle-build-action@v3.5.0
       - name: Build
         run: ./gradlew build
+      - name: Generate OpenAPI
+        run: ./gradlew generateOpenAPI
```
MikeNeilson, line 22: Are we sure we want to do this?

Given the varied implementations it's not likely to be identical between them. Thus is shouldn't be generated at build time.


Then again maybe we can decide there's a single canonical representation of things for the api.

@agilmore2 thoughts?

doc-source/Swagger-README.md
adamkorynta: no that's just a rendering in IntelliJ

adamkorynta:
Missing delete of the manually curated file

.github/workflows/publish.yml
adamkorynta: This provides a good spot to codegen the canonical versions. Generating implementation specific would likely be manually curated anyways, but this would get you 95% of the way there. I would even go so far as to hosting the swagger page as a GitHub Pages site even knowing you couldn't actually test out the queries on a live server, but it still is easier to read than json or yml.

opendcs-integration-test/src/test/java/org/opendcs/odcsapi/owasp/OwaspZap.java
```diff
@@ -0,0 +1,137 @@
+/*
```
adamkorynta, line 1: Why is this class marked new?

opendcs-integration-test/src/test/java/org/opendcs/odcsapi/res/it/OdcsapiResourceIT.java
```diff
@@ -42,7 +42,7 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
-@Tag("integration")
+@Tag("integration-opentsdb-only")
```
adamkorynta, line 45: Shouldn't this be both now?

opendcs-integration-test/src/test/java/org/opendcs/odcsapi/owasp/OwaspZap.java
zack-rma: Looks like this class was duplicated in one of my rebases. I've removed this duplicate class.

opendcs-integration-test/src/test/java/org/opendcs/odcsapi/res/it/OdcsapiResourceIT.java
zack-rma: Correct, I've fixed this.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java
```diff
@@ -0,0 +1,1544 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package org.opendcs.odcsapi.res;
+
+public class ResourceExamples
```
adamkorynta, line 18: All these classes should be final, and I think package private instead of public. empty constructors should have either the comment `//no-opped - utility class` or `throw new AssertionError("Utility class");`

opendcs-rest-api/src/test/resources/org/opendcs/odcsapi/res/it/CWMS/algorithm_ref_add_to_previous.json
adamkorynta, line 1: Why was this file added?

opendcs-rest-api/src/test/resources/org/opendcs/odcsapi/res/it/CWMS/algorithm_tempstring_check.json
adamkorynta, line 1: Why was this file added?

adamkorynta (changes requested):

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ResourceExamples.java
zack-rma: Fixed per suggested changes

opendcs-rest-api/src/test/resources/org/opendcs/odcsapi/res/it/CWMS/algorithm_ref_add_to_previous.json
zack-rma: Removed file, looks like it was added during merge conflict resolution

opendcs-rest-api/src/test/resources/org/opendcs/odcsapi/res/it/CWMS/algorithm_tempstring_check.json
zack-rma: Removed file, looks like it was added during merge conflict resolution

opendcs-rest-api/src/main/webapp/swaggerui/swagger-initializer.js
adamkorynta: More like the answer is both. Zack added a gradle task to generate locally so that you don't have to spin up the web service to create the openapi during development. At runtime though there is an endpoint to query which is provided by the `OpenApiResource` class.

adamkorynta (approved):

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiPropSpec.java
```diff
@@ -15,17 +15,30 @@
...
+	 */
+	@Schema(description = "The name of the property in a Properties object.", example = "debugLevel")
 	private String name = null;
-	
+
 	/**
-	 * A coded string describing the type of the property (see constant prop types herein).
+	 * A coded string describing the type of the property (see constant prop types in the Toolkit PropertySpec class).
 	 */
+	@Schema(description = "A coded string describing the type of the property. Valid values are i (int), n (number)," +
```
MikeNeilson, line 32: should be an enum for rendering in swagger (and any downstream generators)

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiTimeSeriesIdentifier.java
```diff
@@ -15,15 +15,27 @@
...
 
+import io.swagger.v3.oas.annotations.media.Schema;
 
+@Schema(description = "Represents a time series identifier including its unique string, key, description, storage units, and active status.")
 public final class ApiTimeSeriesIdentifier
 {
+	@Schema(description = "A unique string identifying the time series.", example = "OKVI4.Stage.Inst.15Minutes.0.raw")
 	private String uniqueString = null;
+
+	@Schema(description = "The unique numeric key for the time series.", example = "1")
```
MikeNeilson, line 26: This shouldn't be returned to the client. (don't need to fix for this PR, but we should open an issue to track.)

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiTimeSeriesSpec.java
```diff
@@ -17,39 +17,69 @@
...
 	private String statCode = null;
+	@Schema(description = "The interval of time between data points in the time series.", example = "30Minutes")
 	private String interval = null;
+	@Schema(description = "The duration of time that each time series value represents.", example = "0")
 	private String duration = null;
+	@Schema(description = "Version identifier for the time series.", example = "raw")
 	private String version = null;
-	
+
+	@Schema(description = "The unique numeric identifier for the site associated with the time series.", example = "6")
```
MikeNeilson, line 41: Same as the timeseries, but these should actually just the data. Or a least a URL instead of the ID.... though we're already doing a lot of gets on ID aren't we?

MikeNeilson: Arguably they should at least be DbKey instead of Long. There are some open issues/discussions considering moving those to uuid: https://github.com/opendcs/opendcs/issues/625

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/DecodeRequest.java
```diff
@@ -15,29 +15,38 @@
...
+
 /**
  * Used for POST decode payload. Contains the ApiRawMessage and the ApiPlatformConfig
  * to use for decoding.
- * @author mmaloney
  *
+ * @author mmaloney
  */
+@Schema(description = "DTO for decoding requests. Contains raw message data and platform configuration used for decoding.")
 public final class DecodeRequest
```
MikeNeilson, line 27: I guess that answers my question on the DecodesTimeseries objects.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/ComputationResources.java
```diff
@@ -63,13 +71,32 @@ public final class ComputationResources extends OpenDcsResource
...
+			responses = {
+					@ApiResponse(responseCode = "200", description = "Success",
+							content = @Content(mediaType = MediaType.APPLICATION_JSON,
+									array = @ArraySchema(schema = @Schema(implementation = ApiComputationRef.class)))),
+					@ApiResponse(responseCode = "404", description = "No computations found matching the filter criteria"),
+					@ApiResponse(responseCode = "500", description = "Internal Server Error")
+			}
+	)
+	public Response getComputationRefs(@Parameter(schema = @Schema(implementation = String.class))
+		@QueryParam("site") String site,
```
MikeNeilson, line 87: The formatting here just seems wonky, would it be better to group these on one line instead of the nesting?

MikeNeilson (changes requested):
My various pontifications about our current design can certainly be ignored for this effort.

I do think the enums I called out should be done. They've been pretty helpful with swagger in CDA to avoid typos, at least from the swagger-ui.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiPropSpec.java
zack-rma: An enum would be great, but, unless I'm mistaken, it wouldn't support the custom String values of the decodes and Java enums.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiConfigSensor.java
```diff
@@ -19,28 +19,46 @@
...
 	private int sensorNumber = 0;
 
+	@Schema(description = "The name assigned to the sensor.", example = "WL")
 	private String sensorName = null;
 
-	/** V=Variable, F=Fixed, U=undefined */
+	/**
+	 * V=Variable, F=Fixed, U=undefined
+	 */
+	@Schema(description = "The sensor's recording mode. Possible values are V (Variable), F (Fixed), or U (Undefined).",
```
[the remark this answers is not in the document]
zack-rma: Replaced with an enum, updated JSON serialization to support this change

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiConfigScript.java
```diff
@@ -18,20 +18,33 @@
...
 {
+	@Schema(description = "The name of the configuration script.", example = "ST")
 	private String name = null;
-	
-	/** U=undefined, A=ascending, D=descending */
+
+	/**
+	 * U=undefined, A=ascending, D=descending
+	 */
+	@Schema(description = "Defines the data order. Valid values are: U (undefined), A (ascending), D (descending).",
```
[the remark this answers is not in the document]
zack-rma: Swapped for an enum

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiTimeSeriesIdentifier.java
zack-rma: Created an issue here: https://github.com/opendcs/rest_api/issues/382

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiAlgoParm.java
```diff
@@ -15,12 +15,21 @@
...
+	 */
+	@Schema(description = "The role name of the parameter, must be unique within an algorithm.", example = "input1")
 	private String roleName;
-	
-	/** The parameter type -- one of the constant codes defined herein. */
+
+	/**
+	 * The parameter type.
+	 */
+	@Schema(description = "The type of the parameter. Can be input (i) or output (o).", example = "i")
```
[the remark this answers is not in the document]
zack-rma: Updated language to reflect alternative options

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/beans/ApiTimeSeriesSpec.java
MikeNeilson: We'll want to do that throughout, but I wouldn't for this specific effort.

MikeNeilson (approved):

PR #371, opened 2025-03-14
cleanup transitive toolkit dependencies for java ee

adamkorynta:
these dependencies conflict with tomcat dependencies and are causing JSP's not to get properly compiled

## Problem Description

Using transitive dependencies from the toolkit are causing JSP's not to load as they conflict with Tomcat libraries.

## Solution

Exclude unnecessary dependencies.

## how you tested the change

Tested the GUI through the `./gradlew run` task and automated integration tests.

## Where the following done:


opendcs-integration-test/build.gradle
```diff
@@ -19,39 +19,35 @@ plugins {
...
-    implementation(libs.rest.assured)
-    implementation(libs.jersey.container.servlet)
-    implementation(libs.webcompere.system.stubs.jupiter)
     implementation(libs.bundles.tomcat)
     implementation(libs.commons.io)
-    implementation(libs.jul.to.slf4j)
     implementation(libs.slf4j.api)
-    implementation(libs.opendcs)
+    implementation(libs.opendcs) {
+        exclude group: "org.glassfish.jersey.containers"
```
MikeNeilson, line 26: We should actually fix this at the main project level. These's shouldn't be forced dependencies, they're only needed at runtime and should be provided the environment; e.g. only the install subproject of opendcs (and tests) should consume them.

adamkorynta: ah ok, I saw them marked as implementation in the main project

MikeNeilson: Yeah, that's my oops. Still getting used to api vs implementation

adamkorynta: Moved the components to an lrgs-web jar: https://github.com/opendcs/opendcs/pull/983

zack-rma (approved):

MikeNeilson (approved):

PR #372, opened 2025-03-14
fix sonar usage so that it can be run from forked repositories

adamkorynta:
## Problem Description

Running sonar analysis from a forked repository with a user who does not have write permissions fails authentication to the SonarCloud service.

## Solution

Add a github action hook that triggers as the GitHub account to run based on successful build+test execution so that the correct repo permissions are used.

## how you tested the change

Will be testing by getting @zack-rma to PR into this change.

## Where the following done:

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -1,0 +1,53 @@
+name: SonarCloud Scan
+on:
+  workflow_run:
+    workflows: ["Build and Test"]
+    type: [completed]
+
+jobs:
+  Sonar:
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run && github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Set up JDK
+        uses: actions/setup-java@v4.7.0
+        with:
+          java-version: 17
+          distribution: temurin
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4.3.0
+      - name: Checkout PR
+        uses: actions/checkout@v4.2.1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar.yml
```diff
@@ -0,0 +1,53 @@
...
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          run-id: ${{ github.event.workflow_run.id }}
+          path: ./
+          merge-multiple: true
+      - name: Analyze
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: |
+          ./gradlew sonar --info -Dorg.gradle.jvmargs=-Xmx4096M \
```
MikeNeilson, line 49: Also make sure the build.gradle file haven't accidentally set any of this parameters. That bit me a few times in the core project.

adamkorynta: double-checked, the -Xmx is only in the gradle.properties.example so it won't clash. The rest of the properties aren't set

adamkorynta:
huh, looks like I need to get the sonar.yml into main first. I'll wait until feature/task_cwms_support merges first

MikeNeilson:
> huh, looks like I need to get the sonar.yml into main first. I'll wait until feature/task_cwms_support merges first

Yep, but it looks like it's setup correctly at least.

MikeNeilson (approved):

PR #402, opened 2025-04-02
various fixes identified from qa

### LINKED ISSUE #401: Tomcat is no longer responsive if I hit the API frequently over a short period of time.
**Describe the bug**
Using the web GUI, if I spam a bunch of screens over and over, the tomcat instance hangs eventually.

**To Reproduce**
It is not 100% reproducible on the spot, but this is my best method that has worked so far.
1) In the Web GUI, click from screen to screen, which will hit the API a whole bunch of times.
2) Eventually, you will see the screen hang.  If you open the dev tools in chrome (F12 or ctrl-shift-j), you will see on the "Network" tab, that there are a bunch of pending requests.
3) Also, if you load tomcat manager, you will see that even that won't load.  

**Expected behavior**
It is expected that all of these calls return successfully.

**Screenshots**
No screenshots, but here is the jstack - [jstack_bootstrap.txt](https://github.com/user-attachments/files/19572437/jstack_bootstrap.txt)

**Desktop/ServerSide (please complete the following information):**
 - OS of server running tomcat - Windows
 - Version of server running tomcat - 10

**Additional context**
OS of database (PostgreSQL) - Oracle Linux 8

adamkorynta:
## Problem Description

- Last modify times need to be set server-side, not client-side
- tsdb_properties endpoint leaking a connection. Fixes #401 
- missing test coverage on real decodes script and message
- null integer in platform sensor getting incorrectly unboxed

## Solution

- set last modify times server-side
- properly close connection
- add test coverage for CSV message with decodes endpoint
- add null check for boxed integer

## how you tested the change

Tested through web gui and automated integration testing.

## Where the following done:


opendcs-integration-test/src/test/resources/org/opendcs/odcsapi/res/it/DEFAULT/config_input_data.json
```diff
@@ -40,6 +40,10 @@
         }
       }
     ],
-    "formatStatements" : [ ]
+    "formatStatements" : [ {
+      "sequenceNum": 0,
```
MikeNeilson, line 44: should we consider just using the decodesscript test format as the "intermediary representaion"? we do have some plans to allow other script engines for decoding in the future.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PresentationResources.java
```diff
@@ -313,7 +314,7 @@ static PresentationGroup map(DataTypeDAI dai, ApiPresentationGroup presGrp) thro
 			group.parent = parentGroup;
 		}
 		group.dataPresentations = map(dai, presGrp.getElements(), group);
-
+		group.lastModifyTime = new Date();
```
MikeNeilson, line 317: better than nothing, but these should really be database triggers.

opendcs-integration-test/src/test/resources/org/opendcs/odcsapi/res/it/DEFAULT/config_input_data.json
adamkorynta: I think that's a great idea. Probably worth doing after merging the repositories so we can take advantage of the same files.

MikeNeilson (approved):

zack-rma (approved):

PR #409, opened 2025-04-04
add auth to all calls except login/logout

### LINKED ISSUE #189: No API endpoints should be publicly accessible except for login
**Describe the bug**
By allowing public reads of database objects passwords for various things, URLs, and other information will be readable by everything one with access to the API.

Only the Login end point should be accessible publicly, everything else should require at authenticated and authorized user.

adamkorynta:
add integration test that reads all endpoints from the OpenAPI and tests that they all throw 401 unauthorized error codes when the client is not authorized

## Problem Description

Almost (if not all) GET endpoints are public, which can potentially leak sensitive information via properties.

Fixes #189. 

## Solution

Add required authorization to all endpoints except /credentials (this is the login) and /logout

## how you tested the change

Created an integration test that scrapes the OpenAPI (which is auto generated by finding all registered endpoints) and makes a request against them for all available methods (GET/POST/DELETE/etc) and asserts that an unsessioned client will get an error code 401 response.

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AlgorithmResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AlgorithmResources.java
@@ -188,7 +188,7 @@ public final class AlgorithmResources extends OpenDcsResource
 	@Path("algorithm")
 	@Consumes(MediaType.APPLICATION_JSON)
 	@Produces(MediaType.APPLICATION_JSON)
-	@RolesAllowed({ApiConstants.ODCS_API_ADMIN, ApiConstants.ODCS_API_USER})
+	@RolesAllowed({ApiConstants.ODCS_API_USER, ApiConstants.ODCS_API_ADMIN})
 	@Operation(
 			summary = "Create or Overwrite Existing OpenDCS Algorithm",
 			description = "It takes a single OpenDCS Algorithm Record in JSON format, as described above for GET.   \n\n"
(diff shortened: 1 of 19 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AlgorithmResources.java
```diff
@@ -67,7 +67,7 @@ public final class AlgorithmResources extends OpenDcsResource
 	@GET
 	@Path("algorithmrefs")
 	@Produces(MediaType.APPLICATION_JSON)
-	@RolesAllowed(ApiConstants.ODCS_API_GUEST)
+	@RolesAllowed({ApiConstants.ODCS_API_USER, ApiConstants.ODCS_API_ADMIN})
```
wjonassen, line 70: There are some instances where ODCS_API_USER comes before ODCS_API_ADMIN, then there are some in reverse.

Not a big deal, but might as well standardize for consistency.  

Maybe USER then Admin across the board?

zack-rma (changes requested):
Looks like there's one failing test due to the test call not being properly authenticated:

`AlgorithmResourcesIT > getAlgorithmRefs() > OpenDCS-Postgres FAILED
    java.lang.AssertionError: 1 expectation failed.
    Expected status code is <200> but was <401>.`
    
Otherwise these changes look good.

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/AlgorithmResources.java
adamkorynta: updated with a replace all

wjonassen (approved):

zack-rma (approved):

PR #434, opened 2025-04-07
fix missing platform sensors on retrieval

### LINKED ISSUE #412: Platform Sensors are not updated when saving a platform.
**Describe the bug**
When saving a platform (POST), if you add details to the platformSensors key, the details are not saved.

**To Reproduce**
Here is the full json of what I was trying to save, looks like everything saved except for the details of the platformSensor with sensorNum=1
{
    "agency": "NWP_USGS",
    "configId": 9,
    "description": "WILSON RIVER NEAR TILLAMOOK, OR",
    "designator": "goes11",
    "production": false,
    "siteId": 112,
    "platformSensors": [
        {
            "sensorNum": 1,
            "sensorProps": {
                "ADAPS_Medium_Type": "iridium"
            },
            "min": "1",
            "max": "13",
            "usgsDdno": "12"
        },
        {
            "sensorNum": 2,
            "sensorProps": {}
        },
        {
            "sensorNum": 3,
            "sensorProps": {}
        },
        {
            "sensorNum": 4,
            "sensorProps": {}
        },
        {
            "sensorNum": 5,
            "sensorProps": {}
        },
        {
            "sensorNum": 6,
            "sensorProps": {}
        },
        {
            "sensorNum": 7,
            "sensorProps": {}
        },
        {
            "sensorNum": 8,
            "sensorProps": {}
        },
        {
            "sensorNum": 9,
            "sensorProps": {}
        },
        {
            "sensorNum": 10,
            "sensorProps": {}
        }
    ],
    "transportMedia": [
        {
            "assignedTime": 2500,
            "baud": 0,
            "channelNum": 164,
            "dataBits": 0,
            "doLogin": false,
            "loggerType": null,
            "mediumId": "DD67E444",
            "mediumType": "goes-self-timed",
            "parity": "U",
            "password": null,
            "scriptName": "ST",
            "stopBits": 0,
            "timeAdjustment": 0,
            "timezone": "UTC",
            "transportInterval": 3600,
            "transportWindow": 10,
            "username": null
        }
    ],
    "properties": {
        "debugLevel": "33",
        "ignoreSeason": "autumn2",
        "gewagwea": "hewahewa",
        "processSeason": "summer"
    },
    "platformId": 9
}

**Expected behavior**
Those sensor details would be saved.

adamkorynta:
## Problem Description

Fixes #412 . 

## Solution

add missing mapping from PlatformSensors to ApiPlatformSensors

## Where the following done:

### DIFF
--- a/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
+++ b/opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Vector;
+import java.util.stream.Collectors;
 import javax.annotation.security.RolesAllowed;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.Consumes;
@@ -74,6 +75,8 @@ import org.opendcs.odcsapi.errorhandling.DatabaseItemNotFoundException;
 import org.opendcs.odcsapi.errorhandling.MissingParameterException;
 import org.opendcs.odcsapi.errorhandling.WebAppException;
 import org.opendcs.odcsapi.util.ApiConstants;
+
+import static java.util.stream.Collectors.toList;
 
 @Path("/")
 public final class PlatformResources extends OpenDcsResource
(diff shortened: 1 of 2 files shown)


opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
```diff
@@ -272,9 +275,25 @@ static ApiPlatform map(Platform platform)
...
 
+	private static ApiPlatformSensor map(PlatformSensor ps)
+	{
+		ApiPlatformSensor retval = new ApiPlatformSensor();
+		retval.setSensorNum(ps.sensorNumber);
+		if (ps.site != null)
+		{
+			retval.setActualSiteId(ps.site.getId().getValue());
+		}
+		retval.setUsgsDdno(ps.getUsgsDdno());
```
wjonassen, line 292: I don't think this is a required value, ddno.  
Will this fail if it's not there?

wjonassen (changes requested):

opendcs-rest-api/src/main/java/org/opendcs/odcsapi/res/PlatformResources.java
adamkorynta: good catch, this would come back as 0 since it would be the default. Updated to ignore 0

wjonassen (approved):

PR #466, opened 2025-06-01
Improve default behavior for GUI

MikeNeilson:
…tly.

## Problem Description

When attempting to use the GUI in a development environment it took me a minute to realize I need to manually go to /login.jsp and then after login .jsp wasn't added to any links.

## Solution

Implement index.html page that immediately redirects to /portal/login so the end user doesn't need to know that's required.

## how you tested the change

Manually

## Where the following done:

### DIFF
--- a/opendcs-integration-test/src/test/resources/tomcat/conf/context.xml
+++ b/opendcs-integration-test/src/test/resources/tomcat/conf/context.xml
@@ -30,4 +30,5 @@
               maxActive="10"
               maxIdle="4"
               minIdle="1"/>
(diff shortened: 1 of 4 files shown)


adamkorynta:
1. My understanding is that the design was originally wanting to follow the JSP MVC pattern, but then ended up not using the "J" part of JSP and most everything is written in vanilla javascript. @wjonassen can chime in more if I'm missing something.
2. Agreed!

MikeNeilson:
Okay, that did in fact behave better. Was able to get thing started and create a site at least.

PR #553, opened 2025-09-25
Feature/348 container images

MikeNeilson:
Provide docker compose for Postgres implementation to help get people started.
Will also create one for CWMS and the OpenDCS Oracle impl later.

... will also change the branch name later so it makes sense... it's late.


docker-compose.pg.yaml
krowvin, line 1: Can I cry lazy on this one (opinion)

Don't make folks specify a docker compose file with:
`docker compose -f docker-compose.pg.yaml up`

My opinion would be to create a standard docker file for testing and then an opt in one for additional tests. I imagine you eventually want one for oracle too.

krowvin (approved):
Would also appreciate a `.gitattributes` with 
`*.sh text eol=lf`

It will save me on Windows with the IDE defaulting to CLRF without it!

PR #556, opened 2025-09-26
Initial javascript dependency cleanup.

MikeNeilson:
## Problem Description

Fixes #. 

Describe the problem you are trying to solve.

## Solution

- Setup all JavaSript dependencies as a https://Webjars.org dependency in gradle
- Remove all javascript and css that required payment for usage or license otherwise does not allow our usage.

## how you tested the change

Verified that limited behavior was valid:
1. Login works
2. Network tab of developer tools shows no 404 on required dependencies

As limitless was used for the sidebar and other elements, nothing currently renders correctly. This is intentional though I will attempt at least a basic replacement in this PR.

However, given the legal issues, I would rather merge this in with the warts that continue to otherwise provide/release something for which we likely do not hold the legal right to do so.

## Where the following done:


MikeNeilson:
Okay, there's definitely still several problems with the rendering due to the various removed css. Getting the library defaults back in-place helped with several elements but there is a lot that's still not right.

The sidebar especially. A boostrap sidebar extension was added which did help a bit, most of the css class names were valid but several things were removed and it wouldn't surprise if if some behavior was different from what was there.

To be blunt, it's not useable as-is, but only libraries and css we can legally redistribution is now included so that's a better overall situation as it's just the work to fix it.

I'm going to continue on today seeing if I can do that, but otherwise will merge in and just get back to it at some other time or others can pitch in.

MikeNeilson:
Comments and PRs to this branch are both welcome. I will likely merge in over the weekend so there's time to get some initial things in. (currently tweaking the top navbar to newer bootstrap standards now. But will be done, working or not, in the next few minutes after typing this.)

PR #557, opened 2025-09-27
Set JDK 21 as minimum

MikeNeilson:
## Problem Description

Fixes #. 

target branch of main project will no longer support jdks < 21.
## Solution

changed jdk matrices to 17 and 21 only, run integration tests for both.

## how you tested the change

PR build should test everything.

## Where the following done:


MikeNeilson:
NOTE: need to upgrade to gradle 9.2 before any Java 25 builds would work. Considering that may be extensive, it will be handled in a followup PR.

PR #558, opened 2025-09-28
Devops/jsp file org

MikeNeilson:
## Problem Description

Fixes #. 

Reorganize JSP files to provide more clear organization.
Remove implication that project will continue the "DECODES vs TSDB" database split (we're not)

## Solution

- Rename/move files. Changed some to jspf and include `<%@include` directive which is handled at compile time.
- Will also remove "portal" from the path. Looking at the code more I see the value in that we can set certain parameters; however having everything at a sub-path is sub-optimal, we rearrange things so that the same behavior is achieved from /.

NOTE: clearly more work to do, opened as draft for comment.

## how you tested the change

Manually.

## Where the following done:

### DIFF
--- a/opendcs-web-client/build.gradle
+++ b/opendcs-web-client/build.gradle
@@ -27,8 +27,15 @@ dependencies {
     implementation(libs.slf4j.jdk14)
 
     runtimeOnly("org.webjars:bootstrap:5.3.8")
+    constraints {
+        runtimeOnly("org.webjars:jquery") {
+            version {
+                strictly '3.4.1'
+            }
+        }
+    }
     runtimeOnly("org.webjars:jquery-blockui:2.70-1")
-    runtimeOnly("org.webjars:jquery:3.4.1")
+    runtimeOnly("org.webjars:jquery")
     runtimeOnly("org.webjars:select2:4.0.13")
     runtimeOnly("org.webjars:d3js:6.7.0")
     runtimeOnly("org.webjars:uniform:2.1.2-1")
@@ -46,14 +53,30 @@ dependencies {
     runtimeOnly("org.webjars:webjars-servlet-2.x:1.6")
 
     // DataTables + Bootstrap 5
-    implementation("org.webjars.npm:datatables.net:1.13.8")
-    implementation("org.webjars.npm:datatables.net-bs5:1.13.8")
-    implementation("org.webjars.npm:datatables.net-responsive:2.5.0")
-    implementation("org.webjars.npm:datatables.net-responsive-bs5:2.5.0")
-    implementation("org.webjars.npm:datatables.net-buttons:2.4.2")
-    implementation("org.webjars.npm:datatables.net-buttons-bs5:2.4.2")
-    implementation("org.webjars.npm:datatables.net-select:1.7.1")
-    implementation("org.webjars.npm:datatables.net-select-bs5:1.4.0")
+    implementation("org.webjars.npm:datatables.net:1.13.8") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-bs5:1.13.8") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-responsive:2.5.0") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-responsive-bs5:2.5.0") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-buttons:2.4.2") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-buttons-bs5:2.4.2") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-select:1.7.1") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
+    implementation("org.webjars.npm:datatables.net-select-bs5:1.4.0") {
+        exclude group: 'org.webjars.npm', module: 'jquery'
+    }
 
     // Extras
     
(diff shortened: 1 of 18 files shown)


MikeNeilson:
Okay, things are now otherwise behaving correctly. Can add properties to anything. Can't remember if that was buggy before or after I made all these changes.

But since the primary functionally is present, and the goal is organization, I'm comfortable with this in the current state.

krowvin:
Was able to successfully pull down your changes today and get to a login screen!

I was also able to login. 

Some notes of what I see with a fresh session/browser/cache. 
**(some of these may have been issues before)**

1. login page appears to have no padding (text against the edges)
2. login page has no header/it's missing
3. If you click the "back" button after you login and go back to the login page it asks you to login again. Does not remember you logged in and redirect you to the dashboard. 
4. i see this after I login
   image

   And this in the error logs from that page
   image


  <details>
  <summary>
  View log output for Dev console on `http://localhost:7000/platforms`
  </summary>

  ```
VM498 switchery.js:19 Uncaught ReferenceError: module is not defined
    at VM498 switchery.js:19:1
(anonymous) @ VM498 switchery.js:19Understand this error
switchery.js:19 Uncaught ReferenceError: module is not defined
    at switchery.js:19:1
(anonymous) @ switchery.js:19Understand this error
main.js:2 loaded main.js.
decodes.js:26 decodes.js has been loaded.
decodes.js:30 Running Verify Credentials.
decodes.js:58 Verifying token is valid.
dom_utilities.js:17 dom_utilities.js
switchery.js:58 Uncaught TypeError: this.init is not a function
    at new Switchery (switchery.js:58:45)
    at dom_utilities.js:20:25
    at Array.forEach (<anonymous>)
    at HTMLDocument.<anonymous> (dom_utilities.js:19:11)
Switchery @ switchery.js:58
(anonymous) @ dom_utilities.js:20
(anonymous) @ dom_utilities.js:19Understand this error
opendcs_utilities.js:17 opendcs_utilities.js
helpers.js:26 Loaded helpers.js.
waiting.js:2 Waiting Modal js loaded.
notification.js:2 Notification Modal js loaded.
yesno.js:2 yesNo Modal js loaded.
platforms.js:78 Loaded platforms.js.
datatables.js:713 In constructor for Child Class Properties Table.
datatables.js:18 In super constructor for OpenDcsDataTable.
switchery.js:58 Uncaught TypeError: this.init is not a function
    at new Switchery (switchery.js:58:45)
    at platforms.js:1028:25
    at Array.forEach (<anonymous>)
    at initializeElements (platforms.js:1027:11)
    at HTMLDocument.<anonymous> (platforms.js:82:5)
Switchery @ switchery.js:58
(anonymous) @ platforms.js:1028
initializeElements @ platforms.js:1027
(anonymous) @ platforms.js:82Understand this error
platforms:1 [DOM] Found 2 elements with non-unique id #tcpPassword: (More info: https://goo.gl/9p2vKq) <input id=​"tcpPassword" type=​"password" class=​"form-control" required placeholder=​"Password" disabled>​ <input id=​"tcpPassword" type=​"text" class=​"form-control" required placeholder=​"Password" disabled>​Understand this warning
platforms:1 [DOM] Found 2 elements with non-unique id #tcpUsername: (More info: https://goo.gl/9p2vKq) <input id=​"tcpUsername" type=​"text" class=​"form-control" required placeholder=​"Username" disabled>​ <input id=​"tcpUsername" type=​"text" class=​"form-control" required placeholder=​"Username" disabled>​Understand this warning
decodes.js:68 Token Verified Successfully.
datatables.js:752 At Start.
datatables.js:755 At End.
8helpers.js:898 No scroll body for table.  Exiting now.
```

  </details>
  
5. Clicking profile is off of screen/viewport
image
6. Uses relative imports, while okay, can cause issues if you paste style/script tags into other pages and forget they are relative. Consider using the full path for client-side code
i.e.
`<link href="../resources/css/select.dataTables.css" rel="stylesheet" type="text/css">`
7. Noticed you do have to go up one, if the webapp is hosting from the `WEB-INF` directory should we be placing resources in that? Ties into 6 above, but then it would be `<link href="/resources.."etc`
8. Error on 4 looks to be caused by a missing init function for `switcheryjs`, this library appears to be for creating iOS like radio buttons? Have we considered making these in bootstrap? I.e. add the classes required to change color / radius /etc
Alternatively you can apply a boostrap theme, but this adds another dependency. Best to keep it vanilla on one framework in the event of the maintainers of the theme moving on. But here's a theme library: https://bootswatch.com/


There's are some things I found at a quick 5-10 minute click through. I'd have to really sit down and cross reference what i'm finding with existing issues for it to be helpful to your current efforts. (rule out existing issues). 

I could certainly dig deeper, but an example, would be on the sources page

The Properties table does not extend to the full height. You get a tiny scroll overflow but lots of white space. 

image

could look like this:

Easier way to catch this is to "detach" dev tools if you happen to have those open. But i'm also on a higher resolution monitor. 


The culprit seems to be this line though:
`<div class="dt-scroll-body" style="position: relative; overflow: auto; max-height: 100px;">`

With a max height set to a static value the page is not very responsive. At a minimum if you use a static value you'd want to calculate what the potential height for each row would be and go with that. 

I would instead remove the max height, in this case, or use something relative like `max-height: 50vh;` which would account for different devices and their view ports (setting it to 50% of that viewport). 

I imagine most of these CSS issues are a regression due to the required base library changes you had to make. 

It is worth saying that the changes have already made the site very snappy!

MikeNeilson:
Thanks. Some of those are expected, basically 2-6, I'm intentionally not fixing all of the CSS issues in this PR as it's intended to just reorganize. And some of it is broken due to yanking out limitless so drastically.

But I should be able to adopt several of those suggestions immediately.

for 5... yeah, that one confused me, probably a wrong class just haven't looked into it  (none of those buttons actually work at the moment so not really the biggest deal.)

MikeNeilson:
> 7. Noticed you do have to go up one, if the webapp is hosting from the WEB-INF directory should we be placing resources in that? Ties into 6 above, but then it would be <link href="/resources.."etc

point on WEB-INF, if you look at those they should all be under an `<%@include`, that gets processed at compile time. and HTTP get requests can't directly look at the WEB-INF folder. That's part of why the .jspf pages were moved there. Now only the "portal" exposes things, which is better as we hide the file extension in favor of a simple name. Makes it easier long term to tweak the backend without affecting the frontend.

